PR #1, opened 2021-10-19
Knowledge components model

balsa-sarenac:
Moving the initial KC model implemented in https://github.com/Clean-CaDET/platform/pull/84 to the new codebase. For details please refer to the initial PR and its description.

Opened questions:

- [ ] Should `KCProgress` be a value object?
- [ ] Should we model `AssessmentEvent` as an interface or an abstract class?
- [ ] When retrieving KCs from DB should we also grab subcomponents?
- [ ] What would be better for the long term: KC having its subcomponents or KC having its parent?

TODOs:

- [ ] Transition to FluentAPI https://github.com/altmann/FluentResults


src/Tutor.Core/ContentModel/LearningObjects/ArrangeTasks/ArrangeTaskContainerEvaluation.cs
```diff
@@ -1,6 +1,6 @@
 ﻿namespace Tutor.Core.ContentModel.LearningObjects.ArrangeTasks
 {
-    public class ArrangeTaskContainerEvaluation
+    public class ArrangeTaskContainerEvaluation : AssessmentEvent
```
Luburic, line 3: Take care - each Submission has a paired Evaluation. AnswerEvaluation, ArrangeTaskContainerEvaluation, and ChallengeEvaluation.

Here you are mixing and matching different things with the AssessmentEvent. I think we should remove the assessment event for now and think about implementing it after refactoring the challenges and better understanding the Submission - Evaluation flow and how it connects to the KC.

src/Tutor.Web/Controllers/KnowledgeComponent/KCProfile.cs
```diff
@@ -0,0 +1,13 @@
+﻿using AutoMapper;
+using Tutor.Web.Controllers.KnowledgeComponent.DTOs;
+
+namespace Tutor.Web.Controllers.KnowledgeComponent
+{
+    public class KCProfile : Profile
+    {
+        public KCProfile()
+        {
+            CreateMap<Core.KnowledgeComponentModel.KnowledgeComponents.KnowledgeComponent, KnowledgeComponentDTO>();
```
Luburic, line 10: https://github.com/Clean-CaDET/tutor/blob/master/src/Tutor.Web/Controllers/Learners/LearnerProfile.cs very odd.

Luburic (changes requested):
One comment to address - Removing the assessment event for now. We need a better understanding of both the KLI and our learning objects, which is something we should be able to achieve after you refactor the challenge subsystem and we go over the KLI basics once more.

Luburic (approved):
Looks good.

PR #3, opened 2021-11-12
Refactoring challenge subsystem

balsa-sarenac:
Refactored challenge subsystem to consider `codeSnippetId`, where `codeSnippetId` is a unique identifier of a class or a method. Evaluation now finds the corresponding class/method and then evaluates it (instead of evaluating every class and every method from the submission).
Also implemented possible solution for renamed classes/methods. Each `codeSnippetId` has a list of possible renames which is checked if a class/method is not found in the solution.

There is some duplication in the `EvaluateSubmission` method. I was thinking of maybe turning it into a template method. 
Another option could be to move duplicate code in the parent class and introduce two new methods for `ChallengeFulfilmentStrategy` one for evaluating classes and one for methods, that can be overridden in child classes.

Q: Should we make `codeSnippetId` required?


src/Tutor.Core/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/ChallengeFulfillmentStrategy.cs
```diff
@@ -1,13 +1,24 @@
 ﻿using System.Collections.Generic;
+using CodeModel.CaDETModel;
 using CodeModel.CaDETModel.CodeItems;
 
 namespace Tutor.Core.ContentModel.LearningObjects.Challenges.FulfillmentStrategy
 {
     public abstract class ChallengeFulfillmentStrategy
     {
         public int Id { get; private set; }
+        public string CodeSnippetId { get; private set; }
```
Luburic, line 10: We bring these fields to the parent to adhere to DRY.

src/Tutor.Core/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/MetricChecker/BasicMetricChecker.cs
```diff
@@ -1,69 +1,55 @@
...
         }
 
-        public override HintDirectory EvaluateSubmission(List<CaDETClass> solutionAttempt)
+        public override HintDirectory EvaluateSubmission(CaDETProject solutionAttempt)
         {
-            var challengeHints = GetApplicableHintsForIncompleteClasses(solutionAttempt);
-            challengeHints.MergeHints(GetApplicableHintsForIncompleteMethods(solutionAttempt));
-            return challengeHints;
-        }
+            var metrics = solutionAttempt.GetMetricsForCodeSnippet(CodeSnippetId);
```
Luburic, line 21: We utilize an existing method to reduce Feature Envy and code duplication.

src/Tutor.Core/ContentModel/LearningObjects/Challenges/FulfillmentStrategy/NameChecker/BasicNameChecker.cs
```diff
@@ -15,45 +16,50 @@ public class BasicNameChecker : ChallengeFulfillmentStrategy
...
-                namesUsedInCodeSnippet.Add(member.Signature(), GetMemberNames(member));
+                names = GetNames(solutionAttempt, name);
+                if (names != null) return names;
             }
 
-            return namesUsedInCodeSnippet;
+            throw new Exception($"Solution attempt is missing class/method {CodeSnippetId}");
+        }
+
+        private List<string> GetNames(List<CaDETClass> solutionAttempt, string snippetId)
```
Luburic, line 46: Introduced to simplify logic and reduce code duplication.

src/Tutor.Infrastructure/Database/Repositories/LearningObjectDatabaseRepository.cs
```diff
@@ -43,7 +43,7 @@ public Challenge GetChallenge(int challengeId)
                 .Where(c => c.Id == challengeId)
                 .Include(c => c.Solution)
                 .Include(c => c.FulfillmentStrategies)
-                    .ThenInclude(s => (s as BasicMetricChecker).ClassMetricRules)
+                    .ThenInclude(s => (s as BasicMetricChecker).MetricRanges)
                     .ThenInclude(r => r.Hint)
                 .Include(c => c.FulfillmentStrategies)
                     .ThenInclude(s => (s as BasicMetricChecker).MethodMetricRules)
```
Luburic, line 49: **Important**: Update the model, migrations, and scripts to reflect that we now have a single set of metric rules.

Luburic (changes requested):
One thing to address in light of the last commit.

Luburic (approved):
Looks good.

PR #8, opened 2021-12-04
Changing some routes in KCController.

lukaDoric:


Luburic (approved):
We should pay attention to how this feature develops - at this point it's ok to bundle the unit and KC together, but we will probably segregate the classes later. For now it's ok.

PR #12, opened 2021-12-29
Serve suitable assessment event to learner. Update KcMastery.

lukaDoric:


src/Tutor.Infrastructure/Database/Repositories/Domain/KCDatabaseRepository.cs
```diff
@@ -45,12 +61,33 @@ public List<AssessmentEvent> GetAssessmentEventsByKnowledgeComponent(int id)
                 .ThenInclude(c => c.Elements);
             return query.ToList();
         }
+        
+        public List<AssessmentEvent> GetAssessmentEventsByKnowledgeComponentAndLearner(int knowledgeComponentId,
```
Luburic, line 65: GetAssessmentEventsByKnowledgeComponentAndLearner is a lazy and redundant name as it repeats the information from the param list. Furthermore, it is unclear what it means for an AE to be connected to the learner. Maybe GetAssessmentEventsWithLearnerSubmissions, where instead of KCId we have the name relatedKnowledgeComponentId to be more descriptive? It's not ideal, but it is a step in the right direction.

src/Tutor.Core/LearnerModel/Learners/KnowledgeComponentMastery.cs
```diff
@@ -6,5 +6,17 @@ public class KnowledgeComponentMastery
...
         public int LearnerId { get; private set; }
+
+        public KnowledgeComponentMastery(int knowledgeComponentId, int learnerId)
+        {
+            Mastery = 0.0;
+            KnowledgeComponentId = knowledgeComponentId;
+            LearnerId = learnerId;
+        }
+
+        public void SetMastery(double increment)
```
Luburic, line 17: Not a suitable name as it increases the mastery instead of setting it.

src/Tutor.Core/DomainModel/AssessmentEvents/Submission.cs
```diff
@@ -8,11 +8,18 @@ public abstract class Submission
...
+        
+        public double CorrectnessLevel { get; protected set; }
         public DateTime TimeStamp { get; private set; } = DateTime.Now.ToUniversalTime();
 
         public void MarkCorrect()
         {
             IsCorrect = true;
         }
+        
+        public void SetCorrectnessLevel(double correctnessLevel)
```
Luburic, line 20: Use properties for this.

src/Tutor.Infrastructure/Database/Repositories/Domain/AssessmentEventDatabaseRepository.cs
```diff
@@ -40,5 +40,16 @@ public void SaveSubmission(Submission submission)
             _dbContext.Submissions.Attach(submission);
             _dbContext.SaveChanges();
         }
+
+        public Submission FindSubmissionWithMaxCorrectness(Submission submission)
```
Luburic, line 44: Do we need the whole submission or just the AeId and LearnerId? In this case, the method becomes clearer when you work with these two params instead of a single object that has more information that is needed here.

src/Tutor.Core/InstructorModel/Instructors/DefaultInstructor.cs
```diff
@@ -1,17 +1,93 @@
...
+        public DefaultInstructor(IKCRepository ikcRepository, IAssessmentEventRepository assessmentEventRepository)
         {
+            _ikcRepository = ikcRepository;
+            _assessmentEventRepository = assessmentEventRepository;
         }
 
         public Result<AssessmentEvent> SelectSuitableAssessmentEvent(int knowledgeComponentId, int learnerId)
         {
-            throw new System.NotImplementedException();
+            if (_ikcRepository.GetAssessmentEventsByKnowledgeComponent(knowledgeComponentId).Count == 0) return Result.Fail("There are no assessment events yet.");
```
Luburic, line 24: We have an AERepository and the KCRepository, where it seems that both repositories can return AEs. We should either merge both repositories or more clearly define what can be expected from which repo (and choose suitable names). Which option do you think is better and why?

src/Tutor.Core/InstructorModel/Instructors/DefaultInstructor.cs
```diff
@@ -1,17 +1,93 @@
...
+            var suitableAssessmentEvent = GetAssessmentEventWithoutSubmission(knowledgeComponentId, learnerId);
+            return Result.Ok(suitableAssessmentEvent != null
+                ? _assessmentEventRepository.GetAssessmentEvent(suitableAssessmentEvent.Id)
```
Luburic, line 27: To avoid confusion, maybe we can change the name to "GetDerivedAssessmentEvent" to represent that the children have been joined?

src/Tutor.Core/InstructorModel/Instructors/DefaultInstructor.cs
```diff
@@ -1,17 +1,93 @@
...
+
+        public void UpdateKcMastery(Submission submission, int knowledgeComponentId)
+        {
+            var currentCorrectnessLevel = _assessmentEventRepository
+                .FindSubmissionWithMaxCorrectness(submission)?.CorrectnessLevel ?? 0.0;
+            if (currentCorrectnessLevel > submission.CorrectnessLevel) return;
+
+            var kcMastery = _ikcRepository.GetKnowledgeComponentMastery(submission.LearnerId, knowledgeComponentId);
+
+            var kcMasteryIncrement = 100.0 / _ikcRepository.GetAssessmentEventsByKnowledgeComponent(knowledgeComponentId).Count
```
Luburic, line 40: Instead of loading all AEs to count them, you can create a repository function that simply sends a count SQL query (e.g., return _dbContext.AEs.Count(ae => ae.KCid == kcId). Should be faster.

src/Tutor.Core/InstructorModel/Instructors/DefaultInstructor.cs
```diff
@@ -1,17 +1,93 @@
...
+            var assessmentEvents =
+                _ikcRepository.GetAssessmentEventsByKnowledgeComponentAndLearner(knowledgeComponentId, learnerId);
+            return assessmentEvents.FirstOrDefault(ae => ae.Submissions.Count == 0);
+        }
+
+        private Submission FindSubmissionWithMinCorrectness(int knowledgeComponentId, int learnerId)
+        {
+            var lastSubmissions =
+                FindLastSubmissions(
+                    _ikcRepository.GetAssessmentEventsByKnowledgeComponentAndLearner(knowledgeComponentId, learnerId));
```
Luburic, line 58: Can we move this line of code into the FindLastSubmission method?

src/Tutor.Core/InstructorModel/Instructors/DefaultInstructor.cs
```diff
@@ -1,17 +1,93 @@
...
+            var lastSubmissions = new List<Submission>();
+            assessmentEvents.ForEach(ae =>
+            {
+                var currentLastSubmission = ae.Submissions.First();
+                ae.Submissions.ForEach(sub =>
+                {
+                    var compareResult = DateTime.Compare(currentLastSubmission.TimeStamp, sub.TimeStamp);
+                    if (compareResult < 0)
+                    {
+                        currentLastSubmission = sub;
```
Luburic, line 84: lastSubmission is a better name.

src/Tutor.Web/Controllers/Domain/KCController.cs
```diff
@@ -15,11 +16,13 @@ public class KCController : ControllerBase
...
         private readonly IMapper _mapper;
         private readonly IKCService _kcService;
+        private readonly IInstructor _instructor;
 
-        public KCController(IMapper mapper, IKCService kcService)
+        public KCController(IMapper mapper, IKCService kcService, IInstructor instructor)
         {
             _mapper = mapper;
             _kcService = kcService;
+            _instructor = instructor;
```
Luburic, line 25: The instructor is a domain service, while controllers usually communicate with application services. You can leave it for now, but we should refactor this. We should consider introducing an instructor service or utilizing existing suitable services. The method of this service should load the data required by the instructor, who can then perform the needed assessments without interacting with a repository. Once done, the instructor returns the result and the app. service reacts by saving what needs to be saved and return what needs to be returned.

For now we can leave it as is.

Luburic (changes requested):
Several comments inline.

PR #14, opened 2022-01-24
Feature/excel to sql transformer

Luburic:
Slightly expands the Domain model with the following:

- Units and KCs have a "Code".
- Arrange task elements have "Feedback".

Introduces a utility class to process the excel document for creating Units, KCs, IEs, and most AEs and create an SQL script for our model. The class can be activated using the related test.

TODO:

- Introduce wl (world list, working name) assessment event and support for compiling it from the excel into SQL.


balsa-sarenac (approved):
Left some comments inline, but overall looks good!

PR #22, opened 2022-03-10
Feedback and CorrectnessLevel added to ArrangeTaskEvaluationDto,also …

lukaDoric:
…added endopint for KcMastery.


Luburic (approved):
Looks good.

PR #24, opened 2022-03-14
KcMastery pass, completion, and satisfaction tracking

milana-todorovic:
- Adds KnowledgeComponentPassed event and methods for generating it. The KC is currently considered passed when the mastery is >= 0.9.
- Adds KnowledgeComponentCompleted event and methods for generating it. The KC is currently considered completed when the Learner has submitted at least one answer for each AE.
- Adds KnowledgeComponentSatisfied event and methods for generating it. The KC is considered satisfied when the global ['moveOn'](https://github.com/AICC/CMI-5_Spec_Current/blob/quartz/cmi5_spec.md#moveon) criteria is satisfied.


milana-todorovic:
[Track knowledge component pass, completion, and satisfaction](https://trello.com/c/PsJIWhbD/19-track-knowledge-component-pass-completion-and-satisfaction)

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -54,6 +63,54 @@ public Result<AssessmentEvent> SelectSuitableAssessmentEvent(IAssessmentEventSel
...
+                if (assessmentEvent.Id != aeIdForCurrentSubmission && assessmentEvent.Submissions.Count == 0)
+                    return;
+            }
+
+            Causes(new KnowledgeComponentCompleted()
+            {
+                KnowledgeComponentId = KnowledgeComponent.Id,
+                LearnerId = LearnerId
+            });
+            TrySatisfy();
```
Luburic, line 98: Why does TryComplete call TrySatisfy given that TrySatisfy only cares for IsPassed?

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -10,13 +10,19 @@ public class KnowledgeComponentMastery : EventSourcedAggregateRoot
...
+        public bool IsCompleted { get; private set; }
+        public bool IsSatisfied { get; private set; }
 
         private KnowledgeComponentMastery() { }
 
         public KnowledgeComponentMastery(KnowledgeComponent knowledgeComponent)
         {
             Mastery = 0.0;
             KnowledgeComponent = knowledgeComponent;
+            IsPassed = false;
```
Luburic, line 23: I'm still quite unsure of this decision. What is the justification for tracking each field?

Luburic (changes requested):
A few things to consider.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -12,6 +12,7 @@ public class KnowledgeComponentMastery : EventSourcedAggregateRoot
         public int LearnerId { get; private set; }
         public bool IsPassed { get; private set; }
         public bool IsCompleted { get; private set; }
+        public bool IsSatisfied { get; private set; }
```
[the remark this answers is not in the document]
milana-todorovic: Satisfied is from cmi5 terminology, it basically means that the 'moveOn' criteria for the KC has been met. In our case, currently the 'moveOn' criteria for all KCs is 'Passed' (but I believe we discussed switching it to 'Completed' for the test with students).
Unlocked just means that the Learner can access it (so, all of it's children are Satisfied). Although I do suppose that doesn't really need to be an explicit state since it can be calculated at any time.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
milana-todorovic: Mostly because the concept of whether or not an AU (or KC in our case) is satisfied from cmi5 can depend on both it being Passed and it being Completed. While we are currently using an approach that uses the same 'moveOn' criteria for all KCs, we have discussed changing which one we're actually using pretty soon. So the idea was mostly to make that easier.
Actually, I feel like it might even be worth it to implement the concept of 'moveOn' criteria now, and a proper TrySatisfy method that works based on 'moveOn'. Even if we don't plan to do anything like connect different 'moveOn' values to KCs anytime soon (or ever), doing something like that would allow us to maybe put which one we're currently using into configuration instead of changing code.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
milana-todorovic: Refactored to not store IsCompleted in [cd418f5](https://github.com/Clean-CaDET/tutor/pull/24/commits/cd418f5fd84c3d8b8dcd5bc6b935154c35919621).

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
milana-todorovic: I added a globally configurable 'moveOn' criteria in https://github.com/Clean-CaDET/tutor/pull/24/commits/30388fda5f58516240066dc65cc35977ae617d70.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -57,9 +88,54 @@ public Result<AssessmentEvent> SelectSuitableAssessmentEvent(IAssessmentEventSel
...
+                    KnowledgeComponentId = KnowledgeComponent.Id,
+                    LearnerId = LearnerId
+                });
+                TrySatisfy();
+            }
+        }
+
+        private void TryComplete(bool isCompletedBeforeSubmission)
+        {
+            if (isCompletedBeforeSubmission)
```
Luburic, line 113: Why is it important to check if `isCompletedBeforeSubmission` when we check IsCompleted in the next line?

Luburic (commented):
Noticed one minor issue.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -121,5 +144,10 @@ private void When(KnowledgeComponentCompleted @event)
         {
             IsCompleted = true;
         }
+
+        private void When(KnowledgeComponentSatisfied @event)
```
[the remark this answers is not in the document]
milana-todorovic: Since this is an expected consequence of the design of the Apply and When methods, these code smells will just be ignored.

PR #25, opened 2022-03-14
AE selection tracking, and more data added to AEAnswered event

milana-todorovic:
- Adds AssessmentEventSelected event and refactors existing SelectSuitableAssessmentEvent method to generate it.
- Adds Submission content to AssessmentEventAnswered event.

Test are passing (except for the ExcelToSqlTransformerTests which I'm not able to run properly locally because I don't have the input files).

### DIFF
--- a/src/Tutor.Core/DomainModel/AssessmentEvents/SubmissionService.cs
+++ b/src/Tutor.Core/DomainModel/AssessmentEvents/SubmissionService.cs
@@ -29,11 +29,7 @@ namespace Tutor.Core.DomainModel.AssessmentEvents
 
             var result = knowledgeComponentMastery.SubmitAEAnswer(submission);            
 
-            if (result.IsSuccess)
-            {
-                _kcRepository.UpdateKCMastery(knowledgeComponentMastery);
-                _assessmentEventRepository.SaveSubmission(submission);
-            }
+            _kcRepository.UpdateKCMastery(knowledgeComponentMastery);
 
             return result;
         }
--- a/src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventAnswered.cs
+++ b/src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventAnswered.cs
@@ -1,12 +1,10 @@
 ﻿using Tutor.Core.BuildingBlocks.EventSourcing;
+using Tutor.Core.DomainModel.AssessmentEvents;
 
 namespace Tutor.Core.DomainModel.KnowledgeComponents
 {
     public class AssessmentEventAnswered : DomainEvent
     {
-        public int AssessmentEventId { get; set; }
-        public int LearnerId { get; set; }
-        public bool IsCorrect { get; set; }
-        public double CorrectnessLevel { get; set; }
+        public Submission Submission { get; set; }
     }
 }
--- a/src/Tutor.Core/DomainModel/KnowledgeComponents/KCService.cs
+++ b/src/Tutor.Core/DomainModel/KnowledgeComponents/KCService.cs
@@ -50,7 +50,9 @@ namespace Tutor.Core.DomainModel.KnowledgeComponents
         public Result<AssessmentEvent> SelectSuitableAssessmentEvent(int knowledgeComponentId, int learnerId)
         {
             var knowledgeComponentMastery = _ikcRepository.GetKnowledgeComponentMastery(learnerId, knowledgeComponentId);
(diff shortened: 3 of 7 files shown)


src/Tutor.Core/DomainModel/AssessmentEvents/SubmissionService.cs
```diff
@@ -29,11 +29,7 @@ public Result<Evaluation> EvaluateAndSaveSubmission(Submission submission)
 
             var result = knowledgeComponentMastery.SubmitAEAnswer(submission);            
 
-            if (result.IsSuccess)
-            {
-                _kcRepository.UpdateKCMastery(knowledgeComponentMastery);
-                _assessmentEventRepository.SaveSubmission(submission);
-            }
+            _kcRepository.UpdateKCMastery(knowledgeComponentMastery);
```
Luburic, line 32: I take it we removed the IF since UpdatingKCMastery won't do anything if the result is unsuccessful?

src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventAnswered.cs
```diff
@@ -1,12 +1,10 @@
...
 
 namespace Tutor.Core.DomainModel.KnowledgeComponents
 {
     public class AssessmentEventAnswered : DomainEvent
     {
-        public int AssessmentEventId { get; set; }
-        public int LearnerId { get; set; }
-        public bool IsCorrect { get; set; }
-        public double CorrectnessLevel { get; set; }
+        public Submission Submission { get; set; }
```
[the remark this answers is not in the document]
Luburic: Did you check if this serializes the data correctly, given the submission hierarchy?

src/Tutor.Core/DomainModel/KnowledgeComponents/KCService.cs
```diff
@@ -50,7 +50,9 @@ public Result<List<InstructionalEvent>> GetInstructionalEventsByKnowledgeCompone
         public Result<AssessmentEvent> SelectSuitableAssessmentEvent(int knowledgeComponentId, int learnerId)
         {
             var knowledgeComponentMastery = _ikcRepository.GetKnowledgeComponentMastery(learnerId, knowledgeComponentId);
-            return knowledgeComponentMastery.SelectSuitableAssessmentEvent(_assessmentEventSelector);
+            var result = knowledgeComponentMastery.SelectSuitableAssessmentEvent(_assessmentEventSelector);
+            _ikcRepository.UpdateKCMastery(knowledgeComponentMastery);
```
Luburic, line 54: _ikcRepository should be _kcRepository.

As a side note, I see an issue with the design in that we now have to understand which KCM method generates events and requires a call to the repository. The side effect is really not obvious here, leading me to question why the repository is called...

We should probably research if there is a better way to do this. It's not important presently but long term it can easily lead to issues when we forget to save something like this.

As another side note, we should have tests that check if the appropriate events are stored.

Luburic (changes requested):
A few things to consider.

src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventAnswered.cs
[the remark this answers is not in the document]
milana-todorovic: It does (see EventSerializer).

src/Tutor.Core/DomainModel/AssessmentEvents/SubmissionService.cs
milana-todorovic: The if was there because the Submission shouldn't be added if evaluation fails. I moved adding the Submission to the When method, so KcMastery now handles if it's getting added and where. And calling the UpdateKCMastery is fine either way, since it always keeps itself in a valid state when its methods are called.

src/Tutor.Core/DomainModel/KnowledgeComponents/KCService.cs
milana-todorovic: I definitely agree. I feel like it'll generally be true for KcMastery that calling any of its methods will cause changes, but usually it's a lot more apparent than in this case.
 One quick way to fix this would be to maybe actually separate this into two methods - one stays SelectSuitableAssessmentEvent and finds and returns the AE, and the other is something like RecordAssessmentEventSelection. That makes it obvious that KcMastery has been changed and needs to be saved. Although I do feel like that's kind of odd in a different way, it seems strange to call an object to do something, and then have to notify it that the something happened.

PR #26, opened 2022-03-15
Challenge help tracking

milana-todorovic:
- Adds HelpSought event and method for generating it. 

Test are passing (except for the ExcelToSqlTransformerTests which I'm not able to run properly locally because I don't have the input files).


src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -54,6 +55,21 @@ public Result<AssessmentEvent> SelectSuitableAssessmentEvent(IAssessmentEventSel
             return assessmentEventSelector.SelectSuitableAssessmentEvent(KnowledgeComponent.Id, LearnerId);
         }
 
+        public Result SeekHelpForAssessmentEvent(int assessmentEventId, string helpType)
+        {
+            var assessmentEvent = KnowledgeComponent.GetAssessmentEvent(assessmentEventId);
+            if (assessmentEvent == null)
+                return Result.Fail("No assessment event with ID: " + assessmentEventId);
+
```
milana-todorovic, line 63: I would like to check if the help type is appropriate for the type of AE here, but I'm not really sure what a good way to do that would be (or even what a good way to model help type would be, since the string I used for now obviously isn't really good).

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -77,5 +93,12 @@ private void When(AssessmentEventAnswered @event)
 
             Mastery += kcMasteryIncrement;
         }
+
+        private void When(SoughtHelp @event)
+        {
+            /*
+             * Possibly record how many times help was sought in AeMastery?
+             */
+        }
```
milana-todorovic, line 102: I feel like keeping a count of how many times help was requested for an AE could be useful, it seems like it would be useful for AE selection in cases where multiple AEs have the same mastery.

src/Tutor.Infrastructure/Serialization/EventSerializer.cs
```diff
@@ -5,14 +5,16 @@
...
 
 namespace Tutor.Infrastructure.Serialization
 {
     public static class EventSerializer
     {
         private static readonly IDictionary<Type, string> eventRelatedTypes = new Dictionary<Type, string>()
         {
-            { typeof(AssessmentEventAnswered), "AssessmentEventAnswered"}
+            { typeof(AssessmentEventAnswered), "AssessmentEventAnswered" },
+            { typeof(SoughtHelp), "SoughtHelp" }
```
Luburic, line 17: Maybe we should separate "SoughtHints" from "ViewedSolution"?

src/Tutor.Web/Controllers/Plugin/PluginController.cs
```diff
@@ -16,12 +18,14 @@ public class PluginController : ControllerBase
...
+        private readonly IAssessmentEventHelpService _interactionService;
 
-        public PluginController(ILearnerRepository learnerRepository, ISubmissionService submissionService, IMapper mapper)
+        public PluginController(ILearnerRepository learnerRepository, IMapper mapper, ISubmissionService submissionService, IAssessmentEventHelpService interactionService)
         {
             _learnerRepository = learnerRepository;
-            _submissionService = submissionService;
             _mapper = mapper;
+            _submissionService = submissionService;
+            _interactionService = interactionService;
```
Luburic, line 28: interactionService is a way too generic name.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
Luburic: It makes sense, but I would simplify the design for now. The plugin could be enhanced to show one hint at a time, but this is not something we can currently support (no time to write more detailed hints).

src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventHelp/AssessmentEventHelpService.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        public AssessmentEventHelpService(IAssessmentEventRepository assessmentEventRepository, IKCRepository kcRepository)
+        {
+            _assessmentEventRepository = assessmentEventRepository;
+            _kcRepository = kcRepository;
+        }
+
+        public Result SeekHelp(int learnerId, int assessmentEventId, string helpType)
+        {
+            var assessmentEvent = _assessmentEventRepository.GetDerivedAssessmentEvent(assessmentEventId);
```
Luburic, line 19: Do we need to load the whole assessment event (as it has many joins) or can we add a repository function that retrieves the KcID for the AEid?

src/Tutor.Web/Controllers/Plugin/PluginController.cs
```diff
@@ -40,5 +44,17 @@ public ActionResult<LearnerDto> LoginPlugin([FromBody] LoginDto login)
             if (result.IsFailed) return BadRequest(result.Errors);
             return Ok(_mapper.Map<ChallengeEvaluationDto>(result.Value));
         }
+
+        [HttpPost("challenge/hints")]
+        public void SeekHints([FromBody] ChallengeSubmissionDto submission)
+        {
+            _interactionService.SeekHelp(submission.LearnerId, submission.AssessmentEventId, "ChallengeHints");
```
Luburic, line 51: I would have two different methods to differentiate the events.

Luburic (changes requested):
Comments inline.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
milana-todorovic: I definitely agree, I will leave this method empty for now anyway (since there's really nothing that makes sense to put in it yet). The thought about possibly counting how many times the Learner asked for help was mostly just me trying to figure out if this actually fits into KcMastery and how.

src/Tutor.Web/Controllers/Plugin/PluginController.cs
milana-todorovic: Fixed, I accidentally forgot to change that.

src/Tutor.Infrastructure/Serialization/EventSerializer.cs
milana-todorovic: Separated hints and solution into classes in [b2375d3](https://github.com/Clean-CaDET/tutor/pull/26/commits/b2375d3acc0bada1e63dc0824c47dade4d026d31). That's actually the design I initially had for this, but I switched to the version with HelpType because it seemed simpler and I didn't really see any big advantages in either approach at this point, since the general goal is really just to record the events and the functionality itself isn't particularly developed yet.

src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventHelp/AssessmentEventHelpService.cs
milana-todorovic: I added a method to retrieve KcMastery by the AeId and LearnerId in [20803ba](https://github.com/Clean-CaDET/tutor/pull/26/commits/20803ba0cd8ef34974d76b85ab10d3b6c3060eb6). Using that method in AeHelpService is fine since it doesn't actually use the AE, but in SubmissionService the retrieval had to stay as it was (reverted in [049a7be](https://github.com/Clean-CaDET/tutor/pull/26/commits/049a7bebb1a722fe9695aaebec1dffddcb1dc36a)) because it actually needs the entire AE.

src/Tutor.Infrastructure/Serialization/EventSerializer.cs
Luburic: SoughtChallengeSolution sounds good. I'm not really sure about the ancestor, as viewing the solution essentially transforms the AE (challenge) into an IE (video).

src/Tutor.Core/DomainModel/KnowledgeComponents/AssessmentEventHelp/AssessmentEventHelpService.cs
Luburic: It makes sense to add this Repo class and feel free to introduce some repository refactoring along the way.

Luburic (commented):
I've left a few comments inline.

src/Tutor.Infrastructure/Serialization/EventSerializer.cs
milana-todorovic: From that perspective I definitely agree that it would be good to reconsider them having this exact common ancestor, but I think there's no harm in leaving it as is for now (it achieves the main goal of storing the events, and there isn't any information about ancestors in the stored version anyway). 
I feel like the main issue with designing this right now is that really, other than just storing the events, we have no idea what exactly (if anything) we want to do with the exact information that the user viewed hints or the solution of a challenge. The best solution I see is to make KcMastery have a functionality that just receives events from the outside and stores them, but obviously there has to be some sort of validation for which events to accept. Then we can add any events we want to that infrastructure, and just extract them from there into their own methods if we ever want to add an actual functionality related to those events. But I'm honestly really not sure how to design that to make sure that it's just enough of a free slate, but not too much to a point where it can accept just anything. That's why I went with a skeleton of a 'seek help' functionality for these two events instead, but I would really like to discuss better ways to deal with this on our next sync.

milana-todorovic: One attempt at an alternate design for this is in https://github.com/Clean-CaDET/tutor/pull/29.

PR #29, opened 2022-03-16
Arbitrary AE event tracking

milana-todorovic:
This PR is an attempt to design a functionality that would allow us to record arbitrary events related to AEs, while still validating whether or not the event is suitable for the AE in question (as proposed in https://github.com/Clean-CaDET/tutor/pull/26#discussion_r827821326). The idea is that various interactions with AEs would be added through this if we currently just want to record that they happened, without any change in the domain happening as a consequence of the interaction happening. Specific interactions could later be extracted and implemented separately (similarly to how answering AEs / submission evaluation is implemented) if necessary.

### DIFF
--- a/src/Tutor.Core/DomainModel/AssessmentEvents/ArrangeTasks/ArrangeTask.cs
+++ b/src/Tutor.Core/DomainModel/AssessmentEvents/ArrangeTasks/ArrangeTask.cs
@@ -9,6 +9,12 @@ namespace Tutor.Core.DomainModel.AssessmentEvents.ArrangeTasks
         public string Text { get; private set; }
         public List<ArrangeTaskContainer> Containers { get; private set; }
 
+        public override void ValidateInteraction(AssessmentEventInteraction interaction)
+        {
+            if (interaction is not ArrangeTaskInteraction)
+                throw new ArgumentException("Incorrect interaction supplied to Arrange Task with ID " + Id);
+        }
+
         public override Evaluation EvaluateSubmission(Submission submission)
         {
             if (submission is ArrangeTaskSubmission atSubmission) return EvaluateAT(atSubmission);
@@ -18,7 +24,7 @@ namespace Tutor.Core.DomainModel.AssessmentEvents.ArrangeTasks
         private Evaluation EvaluateAT(ArrangeTaskSubmission atSubmission)
         {
             var evaluations = EvaluateContainers(atSubmission.Containers);
-            var correctness = (double) evaluations.Count(c => c.SubmissionWasCorrect) / evaluations.Count;
+            var correctness = (double)evaluations.Count(c => c.SubmissionWasCorrect) / evaluations.Count;
 
             return new ArrangeTaskEvaluation(Id, correctness, evaluations);
         }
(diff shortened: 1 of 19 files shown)


src/Tutor.Core/DomainModel/AssessmentEvents/AssessmentEventInteraction.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace Tutor.Core.DomainModel.AssessmentEvents
+{
+    public abstract class AssessmentEventInteraction
```
milana-todorovic, line 3: I feel like this name is too generic, but I couldn't think of a better one. We need something that is still generic, but not too much or too little.

src/Tutor.Core/DomainModel/KnowledgeComponents/InteractedWithAssessmentEvent.cs
```diff
@@ -0,0 +1,12 @@
...
+using Tutor.Core.DomainModel.AssessmentEvents;
+
+namespace Tutor.Core.DomainModel.KnowledgeComponents
+{
+    public class InteractedWithAssessmentEvent : DomainEvent
+    {
+        public int LearnerId { get; set; }
+        public int AssessmentEventId { get; set; }
+        public AssessmentEventInteraction Interaction { get; set; }
+    }
```
milana-todorovic, line 11: One thing I dislike about the way I designed this for now is the way that the events look when recorded. 

For example, this is how the event for viewing challenge hints looks for now:
<code>
{"type":"InteractedWithAssessmentEvent","learnerId":1,"assessmentEventId":-5,"interaction": 
{"type":"HintsSought"},"timestamp":"2022-03-16T11:15:21.1638662Z"}
</code>

And this is how the event for viewing challenge solutions looks:
<code>{"type":"InteractedWithAssessmentEvent","learnerId":1,"assessmentEventId":-5,"interaction":{"type":"SolutionSought"},"timestamp":"2022-03-16T11:15:22.4721277Z"}</code>

I feel like it might have been better to make the AssessmentEventInteraction class an event, if we did that the hierarchy of events for AEs would look something like:

aeevents

And if we were adding a functionality that happens when, for example, the SoughtChallengeSolution event does, the event would be moved in the hierarchy to inherit AssessmentEventEvent (better name pending) instead.

src/Tutor.Core/DomainModel/AssessmentEvents/ArrangeTasks/ArrangeTask.cs
```diff
@@ -9,6 +9,12 @@ public class ArrangeTask : AssessmentEvent
         public string Text { get; private set; }
         public List<ArrangeTaskContainer> Containers { get; private set; }
 
+        public override void ValidateInteraction(AssessmentEventInteraction interaction)
+        {
+            if (interaction is not ArrangeTaskInteraction)
+                throw new ArgumentException("Incorrect interaction supplied to Arrange Task with ID " + Id);
```
lukaDoric, line 15: Shouldnt we use Result instead exceptions?

milana-todorovic: I generally based the way this looks like right now entirely off of the methods for evaluating Submissions, which is why it uses ArgumentException. I'm pretty sure I asked about this at an earlier sync (if the current design of Evaluate and Evaluation with exceptions is ok, or if it needs to be refactored in the future), and the conclusion was that it's ok since it's pretty localized.

milana-todorovic:
Closed since we decided to stick with the implementation from https://github.com/Clean-CaDET/tutor/pull/26 for now. Might be revisited in the future.

milana-todorovic:
[Track challenge hints and solution viewing](https://trello.com/c/w5UKuFxN/18-track-challenge-hints-and-solution-viewing)

PR #30, opened 2022-03-22
Session launch and termination with event tracking

milana-todorovic:
- Adds SessionLaunched event and method to generate it.
- Adds SessionTerminated event and method to generate it. 
- Adds SessionAbandoned event.


src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -39,10 +41,43 @@ public KnowledgeComponentMastery(KnowledgeComponent knowledgeComponent)
...
+        }
+
+        public void LaunchSession()
+        {
+            if (HasActiveSession)
+                Causes(new SessionAbandoned()
+                {
+                    KnowledgeComponentId = KnowledgeComponent.Id,
+                    LearnerId = LearnerId
+                });
```
milana-todorovic, line 54: I generated this event now to cover cases where termination didn't happen, and it seems to cover that well enough, but note that this won't actually work correctly in cases where an earlier session is actually active.

For example, say a Learner has a KC open in one browser tab, and there is an active session for that. If a Learner opens the same KC in another tab and a new session is launched, a SessionAbandoned event will be generated for the old session, but the Learner won't know that and will actually still be able to interact through the old session. I feel like handling this properly will require adding Ids to sessions or something like that, and connecting all requests to them, so that they can be rejected in cases like this one.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentMastery.cs
```diff
@@ -155,6 +196,21 @@ protected override void Apply(DomainEvent @event)
...
+        }
+
+        private void When(SessionAbandoned @event)
+        {
+            HasActiveSession = false;
+        }
+
         private void When(AssessmentEventAnswered @event)
         {
             /* TODO: refactor to move the GetMaximumSubmissionCorrectness from AE to KCMastery, or a
```
Luburic, line 216: Can you remove these TODOs from the code? You can leave the comments as is and I'll create a task for enhancing the KCM aggregate after the empirical evaluation, once we analyze the data and understand in which direction the KCM should be moving in.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentSatisfied.cs
```diff
@@ -1,10 +1,6 @@
-﻿using Tutor.Core.BuildingBlocks.EventSourcing;
-
-namespace Tutor.Core.DomainModel.KnowledgeComponents
+﻿namespace Tutor.Core.DomainModel.KnowledgeComponents
 {
-    public class KnowledgeComponentSatisfied : DomainEvent
+    public class KnowledgeComponentSatisfied : KnowledgeComponentEvent
```
Luburic, line 3: We should move all events into an Events namespace (and folder) inside the KC namespace.

src/Tutor.Infrastructure/Database/TutorContext.cs
```diff
@@ -117,6 +117,7 @@ private static void ConfigureKcMastery(ModelBuilder modelBuilder)
             modelBuilder.Entity<KnowledgeComponentMastery>().Ignore(kcm => kcm.IsCompleted);
             modelBuilder.Entity<KnowledgeComponentMastery>().Property(kcm => kcm.IsPassed).HasDefaultValue(false);
             modelBuilder.Entity<KnowledgeComponentMastery>().Property(kcm => kcm.IsSatisfied).HasDefaultValue(false);
+            modelBuilder.Entity<KnowledgeComponentMastery>().Property(kcm => kcm.HasActiveSession).HasDefaultValue(false);
```
Luburic, line 120: Why do you need to set default values? In C# an unassigned bool is set to false by default. Can we remove these 3 lines of code?

src/Tutor.Web/Controllers/Domain/KCController.cs
```diff
@@ -68,13 +68,28 @@ public ActionResult<AssessmentEventDto> GetSuitableAssessmentEvent([FromBody] As
...
         {
             var result = _kcService.GetKnowledgeComponentMastery(int.Parse(User.Claims.First(i => i.Type == "id").Value), knowledgeComponentId);
             if (result.IsSuccess) return Ok(_mapper.Map<KnowledgeComponentMasteryDto>(result.Value));
             return NotFound(result.Errors);
         }
+
+        [HttpPost("knowledge-components/{knowledgeComponentId:int}/session/launch")]
+        public ActionResult LaunchSession(int knowledgeComponentId)
+        {
+            _kcService.LaunchSession(int.Parse(User.Claims.First(i => i.Type == "id").Value), knowledgeComponentId);
```
Luburic, line 83: Can we extract `int.Parse(User.Claims.First(i => i.Type == "id").Value` into a separate method?

Luburic (changes requested):
Several things to address. Comments inline.

As a side note, have we stopped writing tests for event sourcing?

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentSatisfied.cs
milana-todorovic: I agree, I think both approaches are used for event sourcing in general (having a separate package, and putting them directly inside relevant packages) and some materials actually insist on not putting them in a separate package, but it has cluttered up the KC namespace quite a bit.

src/Tutor.Core/DomainModel/KnowledgeComponents/IKCService.cs
```diff
@@ -16,9 +16,13 @@ public interface IKCService
         Result<List<AssessmentEvent>> GetAssessmentEventsByKnowledgeComponent(int id);
 
         Result<List<InstructionalEvent>> GetInstructionalEventsByKnowledgeComponent(int id);
-        
+
         Result<AssessmentEvent> SelectSuitableAssessmentEvent(int knowledgeComponentId, int learnerId);
 
         Result<KnowledgeComponentMastery> GetKnowledgeComponentMastery(int learnerId, int knowledgeComponentId);
+
+        void LaunchSession(int learnerId, int knowledgeComponentId);
```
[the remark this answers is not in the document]
milana-todorovic: Switched to Result in https://github.com/Clean-CaDET/tutor/pull/30/commits/0dff27b7cbcdf12e20a94152ecb911954f4f0e1f.

src/Tutor.Infrastructure/Database/TutorContext.cs
milana-todorovic: Removed in https://github.com/Clean-CaDET/tutor/pull/30/commits/e34f53add5e3f433c8e9e8b6fdd5eb58f6853676.

milana-todorovic: Actually, slight correction, these lines did have a purpose. While an unassigned bool in C# is set to false by default, Npgsql doesn't comply with that when generating scripts, it makes the columns required but without a default value: 
```
CREATE TABLE "KcMastery" (
    "Id" integer GENERATED BY DEFAULT AS IDENTITY,
    "Mastery" double precision NOT NULL,
    "KnowledgeComponentId" integer NULL,
    "LearnerId" integer NOT NULL,
    "HasActiveSession" boolean NOT NULL,
    "IsPassed" boolean NOT NULL,
    "IsSatisfied" boolean NOT NULL,
    CONSTRAINT "PK_KcMastery" PRIMARY KEY ("Id"),
    CONSTRAINT "FK_KcMastery_KnowledgeComponents_KnowledgeComponentId" FOREIGN KEY ("KnowledgeComponentId") REFERENCES "KnowledgeComponents" ("Id"),
    CONSTRAINT "FK_KcMastery_Learners_LearnerId" FOREIGN KEY ("LearnerId") REFERENCES "Learners" ("Id") ON DELETE CASCADE
);
```
This isn't really an issue for normal use, but I generally prefer to have the table generated properly with default values. 

It practically only matters for the seeding scripts for tests, since they have to explicitly insert these fields if there are no default values. I'm fine with either keeping these lines or correcting the seeding script though.

src/Tutor.Web/Controllers/Domain/KCController.cs
milana-todorovic: Extracted in https://github.com/Clean-CaDET/tutor/pull/30/commits/66be496dd6889d39e8527ac4462d8659e3c9d753. I wasn't quite sure where to put the method since it will be needed in most controllers, so I made it an extension method for ClaimsPrincipal and I put it in Infrastructure.Security.Authorization.JWT for now.

src/Tutor.Core/DomainModel/KnowledgeComponents/KnowledgeComponentSatisfied.cs
milana-todorovic: Events package added in https://github.com/Clean-CaDET/tutor/pull/30/commits/66be496dd6889d39e8527ac4462d8659e3c9d753, this is its current structure: 

image

I also moved the AssessmentEventHelpService to the AssessmentEvents package, it seemed like a more appropriate place for it since that's where SubmissionService is.

milana-todorovic:
As for tests, I added integration tests for the endpoints from this PR in https://github.com/Clean-CaDET/tutor/pull/30/commits/f2795535144074cfae379736e259455ce8df180b, but tests for if events are generated when they should be still need to be added. I wasn't quite sure how to formulate asserts for that in existing integration tests without it being odd since event generation is basically a side effect and can't be checked for in results of operations. Does something like just querying the event store and checking that it contains the correct events seem alright in this case?

I also think it would be nice to add unit tests for KCM and check event generation there as well, but setting up data for that correctly might be a bit of an issue.

Luburic (approved):
Looks good.

src/Tutor.Core/DomainModel/KnowledgeComponents/KCService.cs
```diff
@@ -67,5 +67,21 @@ public Result<KnowledgeComponentStatistics> GetKnowledgeComponentStatistics(int
             return Result.Ok(new KnowledgeComponentStatistics(
                 mastery.Mastery, assessmentEvents.Count, countCompleted, countAttempted, mastery.IsSatisfied));
         }
+
+        public Result LaunchSession(int learnerId, int knowledgeComponentId)
+        {
+            var knowledgeComponentMastery = _kcRepository.GetKnowledgeComponentMastery(learnerId, knowledgeComponentId);
+            var result = knowledgeComponentMastery.LaunchSession();
```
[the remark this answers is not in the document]
milana-todorovic: Better name for the LaunchSession method, or for the result variable? I'm not sure I really see a better one for either.

PR #39, opened 2022-04-06
User is able to save/update/delete notes.

lukaDoric:


src/Tutor.Core/DomainModel/Notes/NoteService.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public Result<Note> Save(Note note)
+        {
+            return Result.Ok(_noteRepository.Save(note));
+        }
+
+        public Result<int> Delete(int id)
+        {
+            var note = _noteRepository.FindById(id);
```
Luburic, line 22: If this is the only use for FindById why not integrate it into the delete method and remove it from the repo interface?

Luburic (changes requested):
Resolve the codefactor and sonar error along with one minor comment.

Luburic (approved):

PR #43, opened 2022-05-04
Refactor/assessment item mastery

Luburic:


Luburic:
@milana-todorovic I'm introducing the `AssessmentItemMastery` with this PR as we discussed some time ago. When you get the chance please review the changes (possibly in your IDE to get a better sense of the building blocks).

I'm unsure about two things:

1. How should we treat Submissions now? My thinking is that they should be removed from the regular SQL database and handled only as events.
2. Should the KCM lose the direct connection to the KC and instead rely on IDs? This would be a more loosely coupled solution, but we will need to create 2 database queries for some requests (e.g., first check if the KCM exists, then get the related instructional items). This issue can be resolved by introducing a cache for the KC and it's related elements. Since the DomainModel objects will be static during prolonged periods of time and frequently read, it makes sense to cache them.

Luburic:
@milana-todorovic I've made two interesting commits (the last two) that I would like your feedback on.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
```diff
@@ -231,30 +191,32 @@ private void When(SessionAbandoned @event)
...
+                throw new InvalidOperationException("No assessment mastery for item: " + itemId +". Were the masteries created and loaded correctly?");
 
-            var currentCorrectnessLevel = assessmentItem.GetMaximumSubmissionCorrectness();
+            assessmentMastery.LastSubmissionTime = @event.TimeStamp;
+            if (assessmentMastery.Mastery > @event.Submission.CorrectnessLevel) return;
 
-            assessmentItem.Submissions.Add(@event.Submission);
+            assessmentMastery.UpdateMastery(@event.Submission.CorrectnessLevel);
+            UpdateMastery();
+        }
```
milana-todorovic, line 205: We still have a bit of a problem here with where exceptions happen. Since this method doesn't run until the event was already recorded as valid, and it would be used for rehydrating the aggregate (if we ever needed to load the aggregate from events), it should theoretically never fail. So, if there are any exceptions, they need to happen in SubmitAssessmentItemAnswer instead. 
But even if SubmitAssessmentItemAnswer checks that the AssessmentItemMastery exists, this method obviously needs to find it again, and finding an object in a list is an operation that can fail anyway (maybe not really in practice, but theoretically it can). So I'm really not sure how to handle this in a way that is actually formally in line with Event Sourcing conventions, other than having the When method create the AIMastery if it doesn't exist instead of throwing an exception.

src/Tutor.Infrastructure/Database/EventStore/IEventStore.cs
```diff
@@ -1,12 +1,13 @@
...
 using Tutor.Core.BuildingBlocks.EventSourcing;
+using Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries;
 
 namespace Tutor.Infrastructure.Database.EventStore
 {
     public interface IEventStore
     {
-        void Save(EventSourcedAggregateRoot aggregate);
+        void Save(KnowledgeComponentMastery aggregate);
         IEnumerable<DomainEvent> GetEvents(DateTime? start, DateTime? end);
```
milana-todorovic, line 11: I definitely agree with changing the event hierarchy so all events we currently have inherit KnowledgeComponentEvent (I don't actually remember why it wasn't like that in the first place), but I'm not sure why the changes to the event store were necessary to support that? The purpose of the EventStore package and IEventStore especially was really just to serve as a generic facade to hide the technology used to implement the event store, with the idea that it could theoretically be completely extracted from SmartTutor without much changing, so tying it to the actual domain model changes the whole concept quite a bit.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
```diff
@@ -165,9 +128,9 @@ private void TryPass()
...
         }
 
-        private void TryComplete(bool isCompletedBeforeSubmission)
+        private void TryComplete()
         {
-            if (isCompletedBeforeSubmission || !IsCompleted) return;
+            if (IsCompleted) return;
 
             Causes(new KnowledgeComponentCompleted
             {
```
milana-todorovic, line 136: Shouldn't this method check for completion through the AIMasteries before causing the event?

milana-todorovic:
@Luburic 
2. Decoupling KCM from KC seems like a good idea, and so does caching KC and related entities. We just have to make sure KCM objects are constructed carefully to make sure they always match the KC structure, and obviously AIMs have to be constructed at the same time as KCM then (but I'm not sure what to do about https://github.com/Clean-CaDET/tutor/pull/43#discussion_r864987923 if constructing AIMs as needed instead of immediately isn't an option - although constructing them as needed wouldn't be a very good idea anyway because using that when you need all of them for AI selection and such wouldn't look very nice). I'm not sure where UnitEnrollment, KCMs, and AIMs are constructed right now, but it might be a good idea to make the empty constructors for them private just in case (I assume it's not a problem for EFCore since private setters aren't)?

src/Tutor.Infrastructure/Database/EventStore/IEventStore.cs
Luburic: I wanted to create indexes on the LearnerId and KcId fields for the stored domain event. Considering that we are most interested in these types of events for now I found it fitting to repurpose the general DomainEvent.

The question is do we need to treat everything as the same base structure or can we have multiple documents (currently tables) for each major event type?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
Luburic: I can't remember what the issue here was. I've changed back IsCompleted to be a field saved in the database as I found it more convenient to work with for the instructor's analytics. Do I still need the isCompletedBeforeSubmission?

Luburic:
> 

AIMs will be constructed when enrollment happens. They are currently generated through the ExcelToSqlTransformer.

Luburic:
> @Luburic
> 
> 1. Removing Submissions from the regular relational database definitely seems like a good idea, I really can't see any good reason to keep them there and plenty to remove them. With that in mind, I feel like it would be useful to do some more refactoring of the Submission class, the AssessmentItemAnswered event, and related methods:
> 
> * Submission no longer being an entity mean that its properties can be cleaned up. If it no longer needs to be an entity it's mainly a DTO for AssessmentItem's EvaluateSubmission method, so I'd only leave the properties actually needed by that method, which would basically leave the abstract root of the Submission hierarchy empty. That would make the role of Submission more obvious and remove some awkward code, like the part with calling EvaluateSubmission(Submission), getting Evaluation as the return value, and then applying Evaluation to Submission.
> * It's probably also a good moment to rename the EvaluateSubmission to just Evaluate, I think we mentioned that at some point earlier. Speaking of naming, I feel like something like 'Answer' might actually be a slightly more fitting name for the Submission class (The fact that it represent an answer to a question or task seems somewhat more relevant than the fact that it's submitted? Also fits better with the AssessmentItemAnswered event, and the event was named that because I couldn't figure out how to use 'submitted' instead of 'answered' there without it being long and awkward.), but renaming that would be a pretty large change to make for very little payoff.
> * Removing properties from Submission would change the related method signature in ILearnerAssessmentService to something like EvaluateAndSaveSubmission(int learnerId, int assessmentItemId, Submission submission), which is more parameters but actually makes it more consistent with other methods in the service.
> * The AssessmentItemAnswered event would be turned into something like:
> 
> ```
>     public class AssessmentItemAnswered : DomainEvent
>     {
>         public int LearnerId { get; set; }
>         public int AssessmentItemId { get; set; }
>         public Submission Submission { get; set; }
>         public Evaluation Evaluation { get; set; }
>     }
> ```
> 
> That means we can finally move the LearnerId and AssessmentItemId to a common superclass for the events related to the AIs. I also prefer having the Evaluation separate from Submission like this in the event, since I feel like most queries will probably use just one or the other and this way access is more intuitive. Changing the event does mean that the existing data needs to be transformed at some point, but that doesn't seem too hard to do, and can also be postponed by keeping the previous version of the event mapped as well.
> 
> * The one somewhat awkward consequence of this is that KnowledgeComponentMastery's SubmitAssessmentItemAnswer method would now need three parameters - the AI's Id, the Submission, and the Evaluation. If we're going for loose coupling between KCM and KC I don't see a way to avoid that other than adding another DTO, but I'm not sure that seems to necessary in this case.

These are all great comments, thanks for the thorough review. P.S. Check slack as I posted a message a few days ago.

src/Tutor.Infrastructure/Database/EventStore/IEventStore.cs
milana-todorovic: If indexing is the goal, separate columns for LearnerId and KnowledgeComponentId aren't actually required, since Postgres is capable of indexing JSONB columns. Those indexes would have to be created using direct SQL though, since Npgsql doesn't have support for them. The main reason that approach wasn't used for TimeStamp is because time conversions in Postgres can't be used in indexes (and partially because time was assumed to be generic for every possible event).

As for whether or not we need to treat all events as the same base structure, it really depends on what we gain by that and what we lose.
When researching event sourcing, I didn't run into any approach that didn't treat events as the same base structure, because the goal is usually to keep them as structureless as possible. They need to have some metadata to make sure they can be associated to an aggregate and that they can be arranged chronologically, but for the actual payload, the idea is that the event store doesn't know or care what it contains. It's possible to arrange them into multiple collections (although I don't really see how that would work with a relational database and EF Core), but all the collections still have the same schema, and it's generally just an internal detail of the event store.

And for the approach used in these commits, where events don't have the same base structure / the event store actually cares about the structure of the payload instead of treating it as a blob, here are some consequences that I see:
- Adding a new aggregate becomes more difficult. Instead of just creating the aggregate with any events, and using the completely generic event store through a repository, it would be necessary to expand IEventStore (and all implementations) with specific methods for saving that aggregate. 
- Swapping database technologies for the event store becomes more difficult.
- Room for error when adding events to an existing aggregate. While it does make sense to have a one root event per aggregate and inherit from that, there's nothing actually forcing anyone adding new events to respect that. The Causes method as it is now will still accept any inheritor of DomainEvent, but the Save method would fail unexpectedly when it tries to cast it. If the approach we're going for is one root event and one collection per aggregate, EventSourcedAggregateRoot would probably need to become <code>EventSourcedAggregateRoot<T> where T : DomainEvent</code>.

So, what we would lose is the convenience that comes from the event store knowing nothing about the payload structure, which I would say is a pretty big loss, but I honestly have no idea how much we gain. Do we have any new information on what kind of queries we'll need the event store to perform, and how fast they'll actually need to be?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
milana-todorovic: IsCompletedBeforeSubmission isn't needed, since the IsCompleted property isn't changed until the KCCompleted event is raised now. But TryComplete needs to check that all AIs actually have Submissions before causing the event (so, between line 133 and line 135).

Luburic:
@milana-todorovic I've resolved most of the things we discussed, but am leaving the segregating of the KCM and KC for future work. Although the backend portion of this task is pretty simple there are some changes that need to be made on the front-end which will take an hour or two more than I have at this point.

Thanks for all the help, your insight was invaluable 🙂 If you get the chance review the last changes to see if there are any additional tweaks you would make.

PR #47, opened 2022-05-27
fix: Fixes issue with database seeding for integration tests

milana-todorovic:
- Makes the event database use a separate schema from the regular database in integration tests. Previously they used the same schema, which meant that EnsureCreated couldn't be used to initialize the Events table, and also that TutorContext had to be initialized before the Events table. 
- Adds the Collection attribute to integration test classes that were missing it. The missing attribute was causing concurrency issues with database initialization.

### DIFF
--- a/tests/Tutor.Web.Tests/Integration/TutorTestFactory.cs
+++ b/tests/Tutor.Web.Tests/Integration/TutorTestFactory.cs
@@ -19,42 +19,29 @@ public class TutorApplicationTestFactory<TStartup> : WebApplicationFactory<Start
         builder.ConfigureServices(services =>
         {
             using var scope = BuildServiceProvider(services).CreateScope();
-            var scopedServices = scope.ServiceProvider;            
-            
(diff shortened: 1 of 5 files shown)


tests/Tutor.Web.Tests/Integration/TutorTestFactory.cs
```diff
@@ -84,4 +70,18 @@ private static string CreateConnectionStringForTest()
         return
             $"Server={server};Port={port};Database={database};User ID={user};Password={password};Integrated Security={integratedSecurity};Pooling={pooling};";
     }
+
+    private static string CreateConnectionStringForEvents()
```
milana-todorovic, line 74: It's really ugly that this method is basically a copy of CreateConnectionStringForTest, but I wasn't sure how to refactor it right now to make it better. 
Also I feel like it would generally be good to separate all environment variables for the event database from environment variables for the regular database, even though most of them would have the same values right now.

Luburic (approved):
Looks good!

PR #49, opened 2022-06-01
feat: Event query interface

milana-todorovic:
- Adds a basic query interface to the event store, with support for filtering by time, and filtering by arbitrary content of the event.
- Refactors the GetKcStatistics method to use the query interface and removes the GetKcEvents method from IEventStore.

I would also like to add some integration tests for IEventStore and IEventQueryable, but I would appreciate some input on what the best way to do that might be. So far I'm leaning towards making mock implementations of IEventSourcedAggregateRoot, DomainEvent, and IEventSerializer, and having each test case consist of first saving the events and then performing queries and checking the results.


src/Tutor.Infrastructure/Database/EventStore/IEventStore.cs
```diff
@@ -1,15 +1,13 @@
...
 using Tutor.Core.BuildingBlocks;
 using Tutor.Core.BuildingBlocks.EventSourcing;
-using Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries.Events;
 
 namespace Tutor.Infrastructure.Database.EventStore
 {
     public interface IEventStore
     {
         void Save(EventSourcedAggregateRoot aggregate);
+        IEventQueryable Events { get; }
```
Luburic, line 10: Should the repository allow this level of flexibility to its clients?
So far the repositories have encapsulated these queries.

Luburic (commented):
One question inline. We can discuss it and the integration test question in half an hour.

src/Tutor.Infrastructure/Database/EventStore/IEventStore.cs
milana-todorovic: This level of flexibility is appropriate for IEventStore because it isn't really part of the repository layer, but rather just a generic abstraction of an event store meant to be used by specific repositories (although in this case a specific repository doesn't exist yet). It's purpose is to fill the role that EF Core has for relational databases, so it only exists because there isn't a .NET standard way to represent functionalities offered by an event store. 
Although for now it's kind of questionable if this additional layer of abstraction will actually turn out to be useful in this project or just an unnecessary complication. It mostly depends on how many different queries on IEventStore we end up needing. If it's just a few then it would be much easier to just get rid of IEventStore and rewrite all the queries directly in case of a technology swap, but if there's a lot then implementing query translation instead seems worth it.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/QueryBuilder.cs
```diff
@@ -0,0 +1,77 @@
...
+using static Tutor.Infrastructure.Tests.TestData.EventStore.Query;
+
+namespace Tutor.Infrastructure.Tests.TestData.EventStore
+{
+    public static class QueryBuilder
+    {
+        public static IEnumerable<Query> BuildAllPossibleQueries(IEnumerable<DateTime> afterParameters,
+                                                                 IEnumerable<DateTime> beforeParameters,
+                                                                 IEnumerable<ConditionParameter> conditionParameters)
+        {
```
milana-todorovic, line 13: The way this method is implemented probably results in way too many test cases, but I couldn't really think of a way to generate queries that results in less test cases, but still covers as many options as possible. The tests still run pretty quickly though, on my end it takes about 4 seconds to run around 1000 queries.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/TestEventSerializer.cs
```diff
@@ -0,0 +1,45 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text.Json;
+using Tutor.Core.BuildingBlocks.EventSourcing;
+using Tutor.Infrastructure.Database.EventStore;
+using Tutor.Infrastructure.Serialization;
+
+namespace Tutor.Infrastructure.Tests.TestData.EventStore
+{
+    public class TestEventSerializer : IEventSerializer
```
milana-todorovic, line 11: Implementing this for the tests made me realize that the way the event serialization is implemented should probably be refactored a bit. It might be good to move the implementation with Dahomey.Json to the EventStore instead to offer a DefaultEventSerializer. 
I'm not sure if I should do that now since it's a relatively small change, or some other time?

tests/Tutor.Infrastructure.Tests/Unit/EventStore/QueryInterfaceTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        private readonly IEventStore _eventStore;
+
+        public QueryInterfaceTests(EventStoreManager eventStoreManager)
+        {
+            _eventStore = eventStoreManager.EventStore;
+        }
+
+        [Theory]
+        [MemberData(nameof(QueryTestData.GetQueries), MemberType = typeof(QueryTestData))]
+        public void After_before_conditions(Query parameters)
```
milana-todorovic, line 19: There should probably be a better way to test different orders of method calls but I couldn't think of one.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/TestEventTypes.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using Tutor.Core.BuildingBlocks.EventSourcing;
+
+namespace Tutor.Infrastructure.Tests.TestData.EventStore
```
milana-todorovic, line 5: I'm not sure what to do about CodeFactor failing because of this, I feel like it's better to keep these classes in one file in this specific case.

milana-todorovic (commented):
I left a few comments with things I was unsure about. Also, I'm not sure how to fix the SonarCloud issues on this PR, the code smells it shows are pretty confusing and don't offer more information.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/QueryBuilder.cs
milana-todorovic: I implemented a version that generated less queries in 90ca58eed68402dfb1b08202045c985b32aa58e3.

tests/Tutor.Infrastructure.Tests/Unit/EventStore/QueryInterfaceTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        {
+            var _expectedResult = parameters.GetExpectedResult(QueryTestData._allEvents);
+
+            var query = _eventStore.Events;
```
Luburic, line 23: Do we need the event store for these tests, especially since we are targeting unit tests?

Maybe just test the PostgresEventQueryable directly?

As a side note, is `PostgresEventQueryable` the best name? I'm not sure if it has any postgres-specific code?

tests/Tutor.Infrastructure.Tests/TestData/EventStore/QueryBuilder.cs
Luburic: Looking at this closer, I find it pretty challenging to wrap my head around. The whole namespace is full of helper classes to test the QueryInterface. Is there a way to simplify the test design?

tests/Tutor.Infrastructure.Tests/Unit/EventStore/EventStoreManager.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Tutor.Infrastructure.Tests.Unit.EventStore
+{
+    public class EventStoreManager : IDisposable
+    {
+        private readonly string _connectionString = $"Server=localhost;Port=5432;Database=event-store-test;User ID=postgres;Password=super;Integrated Security=false;Pooling=true;";
+        private EventContext _eventContext;
+
+        public IEventStore EventStore { get; private set; }
+
+        public EventStoreManager()
```
Luburic, line 16: Wouldn't this make the test an "integration" test?

Luburic (commented):
Several comments inline. Overall I think the test design is overly complex and that we should find a way to enhance its maintainability.

Maybe reorganizing the folder structure might reduce the complexity, by bundling helpers classes with the test classes, but I'm not sure it is enough.

tests/Tutor.Infrastructure.Tests/Unit/EventStore/EventStoreManager.cs
milana-todorovic: I was pretty unsure about how to categorize it, it might make more sense to call it an integration test.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/TestEventSerializer.cs
milana-todorovic: Implemented in 2d432ad1cd6fdc9bad2beab8cffba35c44fa76dd.

src/Tutor.Infrastructure/Database/EventStore/DefaultEventSerializer/AllowedTypesDiscriminatorConvention.cs
```diff
@@ -19,30 +19,26 @@ public class AllowedTypesDiscriminatorConvention<T> : IDiscriminatorConvention w
...
         public ReadOnlySpan<byte> MemberName => _memberName.Span;
 
-        public AllowedTypesDiscriminatorConvention(JsonSerializerOptions options, IDictionary<Type, T> allowedTypes)
+        public AllowedTypesDiscriminatorConvention(JsonSerializerOptions options, IImmutableDictionary<Type, T> allowedTypes)
             : this(options, allowedTypes, "$type")
         {
         }
 
-        public AllowedTypesDiscriminatorConvention(JsonSerializerOptions options, IDictionary<Type, T> allowedTypes, string memberName)
+        public AllowedTypesDiscriminatorConvention(JsonSerializerOptions options, IImmutableDictionary<Type, T> allowedTypes, string memberName)
```
milana-todorovic, line 27: @Luburic When switching to using immutable dictionaries instead of regular ones I ran into a bug that I think is actually an issue with Dahomey.Json that happens in the following circumstances:
1. A hierarchy contains concrete class A, as well as concrete class B which inherits from A. 
2. When registering classes from the hierarchy in the DiscriminatorConventionRegistry, class B is registered before class A. 

If this happens, registering class A won't actually do anything, so serialization of instances of class A will fail. 

I currently fixed this in the AllowedTypesDiscriminatorConvention by actually performing the work necessary to set up the convention in the constructor instead, including the parts which I believe Dahomey.Json expects to happen while registering types instead. But this patch is a bit odd and might be kind of weird to maintain long term. 

Should we maybe look for another Json library to replace Dahomey.Json? Or will situations like class A and class B ever even happen in event hierarchies? I don't think the current CCaDET hierarchy has any actual case like that (only the test classes for serialization tests do).

tests/Tutor.Infrastructure.Tests/Unit/EventStore/EventStoreManager.cs
milana-todorovic: Moved to integration in cc5665d9268535bb310118fb13485b7cf847701a.

tests/Tutor.Infrastructure.Tests/Unit/EventStore/QueryInterfaceTests.cs
milana-todorovic: We don't actually need any other event store functionalities in this test class, I changed it to just have an IEventQueryable field instead in 125b47c6a721df6ecd9e083acde1a46321a70a8f. 
PostgresEventQueryable is Postgres specific since it used the Json field, I don't think any of the code in the PostgresStore package would work with another EFCore driver because of that. 
These tests aren't specific for PostgresEventQueryable though, they would be the same for any implementation of IEventQueryable. I initially wanted to inject the IEventQueryable through test data, but I didn't find a way to do that with proper database seeding and teardown.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/TestEventTypes.cs
milana-todorovic: Fixed by wrapping in static class.

tests/Tutor.Infrastructure.Tests/Unit/EventStore/QueryInterfaceTests.cs
milana-todorovic: Implemented in 125b47c6a721df6ecd9e083acde1a46321a70a8f.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/QueryBuilder.cs
milana-todorovic: The goal was to somehow generate a wide variety of test cases for better coverage since there are many ways to use IEventQueryable and many things that can go wrong, but it did end up way too complex and confusing. I refactored it quite a bit in 125b47c6a721df6ecd9e083acde1a46321a70a8f, it's still somewhat complex but I believe it should be more readable now (and it should also be much easier to add any future IEventQueryable methods to these tests).

src/Tutor.Infrastructure/Database/EventStore/DefaultEventSerializer/AllowedTypesDiscriminatorConvention.cs
Luburic: I think we will always work with the "leaf" children of the hierarchy. However, this is quite a sensitive situation that should be recorded in a comment. Add a few lines of explanation next to the sensitive code and ensure you mention that it is specific to Dahomey.Json.

There are some indications that System.Text.Json might support polymorphic (de)serialization in .NET 7 so I would wait to see if that is the case before looking for another alternative to Dahomey.

src/Tutor.Web/Controllers/Analytics/AnalyticsController.cs
```diff
@@ -49,8 +49,13 @@ public ActionResult<List<KcStatisticsDto>> GetKcStatistics([FromQuery] int unitI
         var unit = _knowledgeUnitRepository.Get(unitId);
         List<int> learnerIds = null;
         if(groupId != 0) learnerIds = _learnerRepository.GetByGroupId(groupId).Select(l => l.Id).ToList();
+        List<int> kcIds = unit.KnowledgeComponents.Select(kc => kc.Id).ToList();
+
+        var eventQuery = _eventStore.Events.Where(e => kcIds.Contains(e.RootElement.GetProperty("KnowledgeComponentId").GetInt32()));
```
Luburic, line 54: Since this controller is concerned with too many responsibilities, I'm thinking what would be a better approach to segregating the code.

The current IEventStore is generic which is good, but maybe we should be thinking about an IKnowledgeComponentEventStore that offers a method that encapsulates lines 54-57, accepting the ids and return a list of events. It's implementation can work with the IEventStore.

What do you think?

src/Tutor.Web/Controllers/Analytics/AnalyticsController.cs
```diff
@@ -104,7 +109,7 @@ public ActionResult<PagedResult<DomainEvent>> GetEvents([FromQuery] int page, [F
     [HttpGet("all-events")]
     public ActionResult<List<DomainEvent>> GetAllEvents()
     {
-        var result = _eventStore.GetAllEvents();
+        var result = _eventStore.Events.ToList();
```
Luburic, line 112: The changes made here seem to have significantly increased the performance. GetAllEvents completes in about a second, while KcStatistics are compiled in less than a second.

I've changed the ImmutableDictionary to a regular dictionary to see if this affects performance in any way and this was not the case.

For GetAllEvents, the change made here essentially reorganizes identical logic, so I am not sure what is responsible for the performance gain.

As a side note, the current export has an issue when compared to the old one (see images below).
**Current**
image
**Older**
image

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCase.cs
```diff
@@ -0,0 +1,108 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Linq.Expressions;
+using System.Text.Json;
+using Tutor.Core.BuildingBlocks.EventSourcing;
+using Tutor.Infrastructure.Database.EventStore;
+
+namespace Tutor.Infrastructure.Tests.TestData.EventStore
```
Luburic, line 9: Namespace name is not consistent with package.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCaseGenerator.cs
```diff
@@ -0,0 +1,61 @@
...
+
+namespace Tutor.Infrastructure.Tests.TestData.EventStore.EventQueryable
+{
+    public static class EventQueryableTestCaseGenerator
+    {
+        private static IEnumerable<object[]> _testCases = new List<object[]>();
+
+        public static IEnumerable<object[]> GenerateTestCases(IEnumerable<QueryParameter> parameters, int amountPerParameter)
+        {
+            if (!_testCases.Any())
```
Luburic, line 14: Can be inverted to reduce nesting.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCase.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public IEnumerable<DomainEvent> GetExpectedResult(IEnumerable<DomainEvent> events)
+        {
+            var result = events;
+            foreach (var parameter in QueryParameters)
+                result = parameter.ApplyToEnumerable(result);
+            return result;
+        }
+
+        public interface QueryParameter
```
Luburic, line 34: IQueryParameter to respect the convention.

Luburic (commented):
A few things to address, mostly minor. I'm quite curious on the performance gain though.

src/Tutor.Web/Controllers/Analytics/AnalyticsController.cs
milana-todorovic: I would probably just encapsulate these lines in a repository instead, that seems like the best place for all code that uses IEventStore to me (the same way any code that uses EF Core is encapsulated in repositories). But I definitely do agree that the AnalyticsController needs some refactoring.

src/Tutor.Infrastructure/EventConfiguration/EventSerializationConfiguration.cs
```diff
@@ -10,11 +9,11 @@
 using Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries.Events.KnowledgeComponentEvents;
 using Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries.Events.SessionLifecycleEvents;
 
-namespace Tutor.Infrastructure.Serialization
+namespace Tutor.Infrastructure.EventConfiguration
 {
     public static class EventSerializationConfiguration
```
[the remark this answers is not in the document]
milana-todorovic: The EventStore namespace doesn't contain any code specific the Tutor events so probably not there, maybe the root of the folder makes more sense? I generally also feel like the current placement of the whole EventStore namespace is confusing though.

src/Tutor.Web/Controllers/Analytics/AnalyticsController.cs
milana-todorovic: I think I figured out what was causing the performance issues, and it's honestly a pretty stupid mistake - the old version of EventSerializer used to get the serialization options as EventSerializationConfiguration.GetEventSerializerOptions() for every single call to Serialize or Deserialize. This means that the entire JsonSerializerOptions object was instantiated and configured once per event, and I definitely didn't notice just how inefficient that is when I initially wrote the code, but look at it now, just configuring the AllowedTypesDiscriminatorConvention includes copying multiple dictionaries, and I don't even know what exactly the Dahomey.Json initialization methods do and how costly they might be. 

I think the export issue is because of the way the serializer for the controllers is configured, I assume the export expects '$discriminator' instead of 'typeDiscriminator'?

Luburic: @lukaDoric can you comment on the export?

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCase.cs
milana-todorovic: Corrected in 02acfa3104adbf25dc1aaf41773e57833916d87c.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCaseGenerator.cs
milana-todorovic: Corrected in 02acfa3104adbf25dc1aaf41773e57833916d87c.

tests/Tutor.Infrastructure.Tests/TestData/EventStore/EventQueryable/EventQueryableTestCase.cs
milana-todorovic: Corrected in 02acfa3104adbf25dc1aaf41773e57833916d87c.

src/Tutor.Infrastructure/Database/EventStore/DefaultEventSerializer/DefaultEventSerializer.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                registry.RegisterType(type);
+            }
+        }
+
+        public DefaultEventSerializer(IImmutableDictionary<Type, string> eventRelatedTypes) : this(eventRelatedTypes, "$discriminator") { }
+
+        public DomainEvent Deserialize(JsonDocument @event)
+        {
+            return JsonSerializer.Deserialize<DomainEvent>(@event, _options);
```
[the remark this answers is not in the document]
milana-todorovic: Corrected in 02acfa3104adbf25dc1aaf41773e57833916d87c.

src/Tutor.Web/Controllers/Analytics/AnalyticsController.cs
milana-todorovic: I changed it back to '$discriminator' in 02acfa3104adbf25dc1aaf41773e57833916d87c, maybe retry the export now? I would try it but I don't know where the export functionality is or how to use it.

src/Tutor.Infrastructure/Database/EventStore/DefaultEventSerializer/AllowedTypesDiscriminatorConvention.cs
milana-todorovic: Added explanatory comments to AllowedTypesDiscriminatorConvention in f5740ce62e3d0b1c4f2f93a908cfa02cb3fbf66a.

PR #56, opened 2022-09-12
feat: Event store configuration 

milana-todorovic:
- Adds EventStoreOptions class and supporting infrastructure, modelled after EF Core's DbContextOptions, which can be used to easily configure the event store for dependency injection or instantiation. 
- Adds methods for creating and deleting the database and for executing raw scripts to IEventStore to make event store setup for tests less dependent on the concrete database technology used to implement it.


Luburic (commented):
The code looks good. However, I would like a bit more justification for this change.

We are introducing an additional set of classes that are mostly simple individually, but that collectively introduce modest complexity. Two questions arise:
- What do we gain exactly from this change down the line and does it justify the added complexity? 
- Will the additional module for configuration management require any changes as the event store's functionality changes?

milana-todorovic:
@Luburic 
> * Will the additional module for configuration management require any changes as the event store's functionality changes?

Some changes in the event store's functionality might require additions to the configuration module, but nothing too complicated. Here's what I would anticipate in the future if the configuration module is added:
- Any implementation of the event store in a specific technology would need to implement an IEventStoreProvider and add an extension method for EventStoreOptions (like the PostgresStoreProvider and UsePostgres extension method for the Postgres implementation of the event store).
- If any new components that need to be provided by the event store's client (like IEventSerializer) are added, the configuration module would need to be expanded to accommodate that. For example, at some point I considered adding interfaces for resolving event types and aggregate types. This really shouldn't happen much though. 
- Any expansions of IEventStore and IEventQueryable, and additions of new interfaces similar to IEventQuerable to provide indexing, permanent projections, aggregate rehydrating, or similar functionalities shouldn't affect the configuration module. 

That's generally all the somewhat probable future changes in the event store I could think of right now, let me know if I missed anything important. 

> * What do we gain exactly from this change down the line and does it justify the added complexity

I definitely agree that it's questionable if this change brings enough value to justify the amount of boilerplate. In my opinion it does improve the readability of the event store configuration quite a bit, but the only case where I see that truly mattering is in case of a technology swap (especially for test setup and teardown).

Luburic:
My main concern is maintainability in the long run, particularly by other devs. The code is smart, which is good for solving the configuration issue, but bad if we need to change it in a year. I'm ok with it changing if we are swapping to a new data store, but I want to avoid requiring configuration changes when we introduce new domain functionality.

Do you think the upcoming PRs (expanding the `EventSourcedAggregate`, introducing the learning session, etc.) will help us better understand if and how this configuration might change? If so, maybe we can leave the PR for now and review it in a few weeks.

milana-todorovic:
@Luburic 
I think leaving this PR open for some time and re-evaluating it later makes sense, since it is questionable if it's something we really need or not. 

I don't think this change has the potential to have a large negative effect on later development though. In general, the configuration module (and the entirety of the event store module) shouldn't be affected at all by domain changes. Any new event classes will need to be added to the dictionary in the <code>EventConfiguration</code> class so they can be serialized, but that's pretty much it (and that was also the case before this PR). 
Changes in <code>BuildingBlocks</code> related to expanding the <code>EventSourcedAggregate</code> with <code>EventSourcedEntity</code> also shouldn't affect the event store or its configuration. The public interface of <code>EventSourcedAggregate</code> with <code>GetChanges</code>, <code>ClearChanges</code>, and <code>LoadFromHistory</code> is the only thing that matters to the event store, and I really don't see that interface changing at all in the future since it's a pretty straightforward component of the event sourcing pattern.

Luburic:
Closing this PR for now but keeping the branch active to return to it at a later time when we have the resources to resolve conflicts and upgrade this feature.

PR #58, opened 2022-09-18
feat: Session duration tracking. Event sourced entities

milana-todorovic:
Adds:

- EventSourcedEntity which can cause events on its EventSourcedAggregateRoot.
- A SessionTracker entity in the KCM aggregate, responsible for session lifecycle management. 

I left some comments with things I'm unsure about.

### DIFF
--- a/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
+++ b/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
@@ -6,11 +6,32 @@ namespace Tutor.Core.BuildingBlocks.EventSourcing
     {
         private EventSourcedAggregateRoot _root;
 
+        /// <summary>
+        /// Initializes the entity for event sourcing. 
+        /// </summary>
+        /// <remarks>
+        /// Override if the entity contains other <see cref="EventSourcedEntity"/> objects which needs to be initialized. 
+        /// This method should only be invoked from 
+        /// <see cref="EventSourcedAggregateRoot.Initialize"/> (for entities contained by the aggregate root)
+        /// or <see cref="Initialize(EventSourcedAggregateRoot)"/> (for entities contained by other entities).
+        /// </remarks>
+        /// <param name="root"></param>
+        /// <exception cref="ArgumentNullException"></exception>
         public virtual void Initialize(EventSourcedAggregateRoot root)
         {
             _root = root ?? throw new ArgumentNullException(nameof(root));
         }
 
+        /// <summary>
+        /// Forwards the event to <see cref="EventSourcedAggregateRoot.Causes(DomainEvent)"/> to be recorded and applied
+        /// to the aggregate.
+        /// </summary>
+        /// <remarks>
+        /// This method should be called from the entity's command methods if invariants are satisfied 
+        /// and the aggregate state needs to be changed.
+        /// </remarks>
+        /// <param name="event">Must be valid in the entity's current state.</param>
+        /// <exception cref="EventSourcingException"></exception>
         protected void Causes(DomainEvent @event)
         {
             if (_root == null)
@@ -18,6 +39,17 @@ namespace Tutor.Core.BuildingBlocks.EventSourcing
             _root.Causes(@event);
         }
 
+        /// <summary>
+        /// Applies changes represented by an event to the entity.
+        /// </summary>
+        /// <remarks>
+        /// Changes to entity state should happen exclusively in this method.
+        /// Implementation should assume that the event is valid and apply changes without checking invariants. 
+        /// This method should only be invoked from 
+        /// <see cref="EventSourcedAggregateRoot.Apply(DomainEvent)"/> (for entities contained by the aggregate root) 
+        /// or <see cref="Apply(DomainEvent)"/> (for entities contained by other entities).
+        /// </remarks>
+        /// <param name="event"></param>
         public abstract void Apply(DomainEvent @event);
     }
 }
--- a/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
+++ b/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
@@ -6,13 +6,15 @@ namespace Tutor.Core.BuildingBlocks.EventSourcing
     {
         private EventSourcedAggregateRoot _root;
 
-        public void Initialize(EventSourcedAggregateRoot root)
+        public virtual void Initialize(EventSourcedAggregateRoot root)
         {
             _root = root ?? throw new ArgumentNullException(nameof(root));
         }
 
         protected void Causes(DomainEvent @event)
         {
+            if (_root == null)
+                throw new EventSourcingException("Entity of type " + GetType() + " not initialized for Event Sourcing.");
             _root.Causes(@event);
         }
 
--- a/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
+++ b/src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
@@ -1,0 +1,21 @@
+﻿using System;
+
+namespace Tutor.Core.BuildingBlocks.EventSourcing
+{
+    public abstract class EventSourcedEntity : Entity
+    {
+        private EventSourcedAggregateRoot _root;
+
+        public void Initialize(EventSourcedAggregateRoot root)
+        {
+            _root = root ?? throw new ArgumentNullException(nameof(root));
+        }
+
+        protected void Causes(DomainEvent @event)
+        {
+            _root.Causes(@event);
+        }
+
+        public abstract void Apply(DomainEvent @event);
+    }
+}
--- a/src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
+++ b/src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
@@ -49,6 +49,7 @@ namespace Tutor.Core.LearnerModel.DomainOverlay
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
+            kcMastery.Initialize();
             var result = kcMastery.GetInstructionalItems();
             _kcMasteryRepository.UpdateKcMastery(kcMastery);
 
@@ -60,6 +61,7 @@ namespace Tutor.Core.LearnerModel.DomainOverlay
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
+            kcMastery.Initialize();
             var result = kcMastery.SelectSuitableAssessmentItemId(_assessmentItemSelector);
             _kcMasteryRepository.UpdateKcMastery(kcMastery);
 
@@ -81,6 +83,7 @@ namespace Tutor.Core.LearnerModel.DomainOverlay
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
+            kcMastery.Initialize();
             var result = kcMastery.LaunchSession();
             _kcMasteryRepository.UpdateKcMastery(kcMastery);
             return result;
@@ -91,6 +94,7 @@ namespace Tutor.Core.LearnerModel.DomainOverlay
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
+            kcMastery.Initialize();
             var result = kcMastery.TerminateSession();
             _kcMasteryRepository.UpdateKcMastery(kcMastery);
             return result;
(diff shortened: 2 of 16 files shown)


src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/Events/KnowledgeComponentEvent.cs
```diff
@@ -1,10 +1,13 @@
...
+using Tutor.Core.BuildingBlocks.EventSourcing;
 
 namespace Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries.Events
 {
     public abstract class KnowledgeComponentEvent : DomainEvent
     {
         public int KnowledgeComponentId { get; set; }
         public int LearnerId { get; set; }
+        public TimeSpan DurationOfPreviousSessions { get; set; }
+        public TimeSpan DurationOfCurrentSession { get; set; }
```
milana-todorovic, line 11: For now I just added these to all KC events as an example, I wasn't sure what to include in which events from the data we now have in SessionTracker. 

Also, the TimeSpan class is serialized like this by default:
image
So that should probably be cleaned up if we keep TimeSpan properties in events, but I'm not quite sure what the easiest way to do that is.

src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
```diff
@@ -0,0 +1,21 @@
...
+namespace Tutor.Core.BuildingBlocks.EventSourcing
+{
+    public abstract class EventSourcedEntity : Entity
+    {
+        private EventSourcedAggregateRoot _root;
+
+        public void Initialize(EventSourcedAggregateRoot root)
+        {
+            _root = root ?? throw new ArgumentNullException(nameof(root));
+        }
```
milana-todorovic, line 12: Since the EventSourcedEntity needs a reference to the EventSourcedAggregateRoot in order to work, I added an Initialize method to EventSourcedAggregateRoot which can then initialize all entities as needed. I wasn't quite sure where the best place to invoke the Initialize method is though, for now I placed it in the services before methods are called on KnowledgeComponentMastery. 

Also, since EventSourcedAggregate and EventSourcedEntity expect pretty specific behavior from inheriting classes, should I add some documentation comments explaining that? Most of it is pretty standard for event sourcing, but the addition of entities complicates things a bit so it might be good to have some extra clarity.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
```diff
@@ -22,11 +22,12 @@ public class KnowledgeComponentMastery : EventSourcedAggregateRoot
     public KnowledgeComponent KnowledgeComponent { get; private set; }
     public List<AssessmentItemMastery> AssessmentMasteries { get; private set; }
     public double Mastery { get; private set; }
-    public bool HasActiveSession { get; private set; }
     public bool IsStarted { get; private set; }
     public bool IsPassed { get; private set; }
     public bool IsSatisfied { get; private set; }
     public bool IsCompleted { get; private set; }
+    public SessionTracker SessionTracker { get; private set; }
+    public bool HasActiveSession { get; private set; }
```
milana-todorovic, line 30: The HasActiveSession property needs to be removed, but to do that, all the seeding scripts need to be corrected. Is there a way to do that without fixing all the inserts for KCM manually?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/SessionTracker.cs
```diff
@@ -0,0 +1,84 @@
...
+        }
+
+        private void When(SessionLaunched @event)
+        {
+            StartOfCurrentSession = @event.TimeStamp;
+            LastActivity = @event.TimeStamp;
+
+            @event.DurationOfPreviousSessions = DurationOfPreviousSessions;
+            @event.DurationOfCurrentSession = DurationOfCurrentSession;
+        }
```
milana-todorovic, line 52: I'm not sure if adding data to events inside the Apply method like I did here and in KCM is generally acceptable in event sourcing, but I feel like it's be preferable to trying to initialize the entire event when it's first instantiated in this specific case, since our events contain a lot of data which will be useful for analytics, but which isn't actually needed for applying changes to the aggregate. It ends up being a bit messy either way, but I feel like the behavior methods are a lot more readable this way.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/SessionTracker.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries
+{
+    public class SessionTracker : EventSourcedEntity
+    {
+        public TimeSpan DurationOfPreviousSessions { get; private set; } = new TimeSpan(0);
+        public TimeSpan DurationOfCurrentSession { get => HasActiveSession ? LastActivity.Value - StartOfCurrentSession.Value : new TimeSpan(0); }
+        public bool HasActiveSession { get => StartOfCurrentSession.HasValue; }
+        public DateTime? StartOfCurrentSession { get; private set; }
+        public DateTime? LastActivity { get; private set; }
```
milana-todorovic, line 15: Also, now that we're tracking the time of the last activity in a session, maybe we could add a simple improvement for session abandonment by implementing a method in the KCM aggregate which abandons the session if it's been inactive for a certain amount of time? Then we can add a scheduled process or something similar that would find all KCM's with active sessions, and abandon them if necessary.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/SessionTracker.cs
```diff
@@ -0,0 +1,84 @@
...
+            DurationOfPreviousSessions += DurationOfCurrentSession;
+            StartOfCurrentSession = null;
+            LastActivity = null;
+        }
+
+        private void When(KnowledgeComponentEvent @event)
+        {
+            LastActivity = @event.TimeStamp;
+
+            @event.DurationOfPreviousSessions = DurationOfPreviousSessions;
```
Luburic, line 80: It would be helpful if I can see a sequence diagram (simplest possible sketch is enough) to demonstrate how the service invokes these events for the more complex design of the event sourced aggregate.

Luburic (commented):
A few minor comments and several questions to discuss.

src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
```diff
@@ -49,6 +49,7 @@ public Result<List<InstructionalItem>> GetInstructionalItems(int knowledgeCompon
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
+            kcMastery.Initialize();
```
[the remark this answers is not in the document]
milana-todorovic: I'm not a fan of this either, but I wasn't sure where else to put it. Maybe inside KCM's methods?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/Events/KnowledgeComponentEvent.cs
```diff
@@ -1,10 +1,13 @@
...
+﻿using System;
+using Tutor.Core.BuildingBlocks.EventSourcing;
 
 namespace Tutor.Core.LearnerModel.DomainOverlay.KnowledgeComponentMasteries.Events
 {
     public abstract class KnowledgeComponentEvent : DomainEvent
     {
         public int KnowledgeComponentId { get; set; }
         public int LearnerId { get; set; }
+        public TimeSpan DurationOfPreviousSessions { get; set; }
```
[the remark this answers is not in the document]
milana-todorovic: The first one is the total duration of all previous sessions for the KC (excluding the current one), so essentially how much time the Learner has spent on interactions with this KC in total. But as I mentioned in https://github.com/Clean-CaDET/tutor/pull/58#discussion_r973745701, these fields are really just an example I used to check if everything is working ok with SessionTracker. We should discuss what data to actually include in events from what we now have available.

src/Tutor.Infrastructure/Database/TutorContext.cs
```diff
@@ -116,6 +116,7 @@ private static void ConfigureKcMastery(ModelBuilder modelBuilder)
             modelBuilder.Entity<KnowledgeComponentMastery>()
                 .HasMany(kcm => kcm.AssessmentMasteries)
                 .WithOne();
+            modelBuilder.Entity<KnowledgeComponentMastery>().OwnsOne(kcm => kcm.SessionTracker).Ignore(tracker => tracker.Id);
```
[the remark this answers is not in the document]
milana-todorovic: To map SessionTracker to the database. Since it has a one to one relationship with KCM, I wanted it to be mapped as part of the same table to avoid unnecessary joins.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/SessionTracker.cs
milana-todorovic: I'm not really sure what to include in the sequence diagram? As far as the service is concerned everything stays the same, it just invokes public command methods on the aggregate root.
Or do you mean a diagram of what a command method of the aggregate root would look like in cases where events are caused by an entity or need to be applied to an entity?

Luburic: Yes, I was aiming for the scenario where the root and the entity are interacting through events.

milana-todorovic: Basically, the only two possible interactions are:

- The aggregate root can delegate some part of a command method to an entity by calling one of the entity's command methods. If the entity method needs to cause an event, that's achieved by calling the Causes method on the root, so events caused by entities are treated the same as events caused by the root.
- When updating the aggregate state based on an event through the root's Apply method, an entity might need to be updated. That's achieved by passing the event to the entity's Apply method. 

Below are sequence diagrams illustrating a simple case of those two interactions:

src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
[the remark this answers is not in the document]
milana-todorovic: Moved call to Initialize to repository in 5cc0ee555d6397b2c2e3a4add432510cd77a4983 as discussed during sync.

src/Tutor.Core/BuildingBlocks/EventSourcing/EventSourcedEntity.cs
milana-todorovic: Initialize moved to repositories in 5cc0ee555d6397b2c2e3a4add432510cd77a4983. Documentation comments added in 4140ef78c68f963478ce26a14c694ec78fa88160.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/Events/KnowledgeComponentEvent.cs
milana-todorovic: Event properties adjusted in e27e7e9274f6a6072fed45df0051c37a95703439.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/Events/KnowledgeComponentEvent.cs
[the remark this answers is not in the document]
milana-todorovic: Event properties adjusted as discussed during sync in e27e7e9274f6a6072fed45df0051c37a95703439.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
milana-todorovic: Removed HasActiveSession and fixed scripts in d082c4384bf7a8ad1a6e3c772af2103fa477decc.

src/Tutor.Infrastructure/Database/DataImport/LearnerSqlExporter.cs
```diff
@@ -72,10 +72,10 @@ private static string BuildEnrollmentSql(List<UserLearnerColumns> learners, Doma
                 foreach (var kcId in domainContent.KnowledgeComponents.Select(kc => kc.Id))
                 {
                     sqlBuilder.Append(
-                        "INSERT INTO public.\"KcMasteries\"(\"Id\", \"Mastery\", \"KnowledgeComponentId\", \"LearnerId\", \"IsStarted\", \"IsPassed\", \"IsSatisfied\", \"IsCompleted\", \"HasActiveSession\") VALUES");
+                        "INSERT INTO public.\"KcMasteries\"(\"Id\", \"Mastery\", \"KnowledgeComponentId\", \"LearnerId\", \"IsStarted\", \"IsPassed\", \"IsSatisfied\", \"IsCompleted\") VALUES");
                     sqlBuilder.AppendLine();
                     sqlBuilder.Append("\t(" + ++kcMasteryId + ", 0.00, " + kcId + ", "
-                                       + learnerId + ", false, false, false, false, false);");
+                                       + learnerId + ", false, false, false, false);");
                     sqlBuilder.AppendLine().AppendLine();
```
milana-todorovic, line 79: The seeding script needs to be generated again because of this change, but I can't do it since I don't have the Excel files.

Luburic: Notes. I will expand the authoring tools to include @lukaDoric `Course` additions on Monday and update the seed-RS and starting templates accordingly.

Luburic (approved):

PR #61, opened 2022-09-25
feat: Get course by id.

lukaDoric:


tests/Tutor.Web.Tests/Integration/Instructors/InstructorTests.cs
```diff
@@ -2,6 +2,7 @@
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.Extensions.DependencyInjection;
 using Shouldly;
+using Tutor.Web.Controllers.Domain.DTOs;
```
Luburic, line 5: Do we need a test for the new endpoint?

Luburic (approved):
Ok in general, but maybe we need a test?

PR #60, opened 2022-09-23
refactor: Converts AssessmentItemMastery into an EventSourcedEntity

milana-todorovic:

PR #62, opened 2022-09-27
refactor: Proposes a different approach for handling AI interactions in KCM. 

milana-todorovic:
After working on #60, I noticed that almost half of <code>KnowledgeComponentMastery</code> methods just record assessment item interactions, and that all of them now have identical basic structure where they just perform the session check, find the appropriate <code>AssessmentItemMastery</code>, and forward the call to it. I feel like this already negatively impacts the readability and maintainability of the KCM aggregate, and while it's not that bad yet, it's pretty likely to become completely unmanageable in the future (since we'll probably want to record even more assessment item interactions, and add instructional item interactions as well). 

This PR proposes the following changes which are meant to address this issue:

- <code>AssessmentItemMastery</code> now exposes only one command method - <code>RecordInteraction</code>, which receives an <code>AssessmentItemEvent</code> and causes it. This method should also validate the event first in the future, but it doesn't have validation right now since the old methods didn't either.
- <code>KnowledgeComponentMastery</code> now exposes one command method for interacting with assessment items - RecordInteractionWithAssessmentItem</code>, which receives an <code>AssessmentItemEvent</code> representing the interaction. This method performs the session check, finds the <code>AssessmentItemMastery</code>, forwards the call to it, and then performs the checks for pass, completion, and satisfaction if necessary.
- Previously checks for pass and completion happened only when recording assessment item answers, but I feel like it actually makes more sense to treat all interactions with assessment items as having the potential to cause pass or completion, especially for completion. Otherwise any change to how <code>AssessmentItemMastery</code> tracks mastery would result in changes to <code>KnowledgeComponentMastery</code> as well. Initially I implemented that by just always calling <code>TryPass</code> and <code>TryComplete</code> when recording assessment item interactions, but just in case that causes a performance hit, I added interfaces for marking events which might cause pass or completion instead. 

I also included some minor renaming changes of <code>KnowledgeComponentMastery</code> methods to make everything consistent.


src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
```diff
@@ -49,23 +51,31 @@ public Result<List<InstructionalItem>> GetInstructionalItems(int knowledgeCompon
             var kcMastery = _kcMasteryRepository.GetFullKcMastery(knowledgeComponentId, learnerId);
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
-            var result = kcMastery.SelectInstructionalItems();
+            var result = kcMastery.RecordInstructionalItemSelection();
```
Luburic, line 54: The name implies that this is a command when it is more of a query with a side-effect that produces an event...
Working with aggregates is hard :)

src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
```diff
@@ -49,23 +51,31 @@ public Result<List<InstructionalItem>> GetInstructionalItems(int knowledgeCompon
...
 
-            var result = kcMastery.SelectAssessmentItem(_assessmentItemSelector);
+            var selectionResult = _assessmentItemSelector.SelectSuitableAssessmentItemId(kcMastery.AssessmentItemMasteries, kcMastery.IsPassed);
+            if (selectionResult.IsFailed) return selectionResult.ToResult<AssessmentItem>();
+
+            var masteryResult = kcMastery.RecordAssessmentItemInteraction(new AssessmentItemSelected()
+            {
+                AssessmentItemId = selectionResult.Value
+            });
+            if (masteryResult.IsFailed) return masteryResult.ToResult<AssessmentItem>();
```
Luburic, line 74: Is making the service smarter smart?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/AssessmentItemMastery.cs
```diff
@@ -58,5 +33,10 @@ private void When(AssessmentItemAnswered @event)
             SubmissionCount++;
             LastSubmissionTime = @event.TimeStamp;
         }
+
+        private void When(AssessmentItemEvent @event)
```
Luburic, line 37: Should the `AssessmentItemEvent` be abstract and not need a standalone When?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
```diff
@@ -70,40 +68,29 @@ private void JoinOrLaunchSession()
...
+        var aim = AssessmentItemMasteries.Find(aim => aim.AssessmentItemId == interaction.AssessmentItemId);
+        if (aim == null) return Result.Fail("No mastery for assessment item with id " + interaction.AssessmentItemId +
+                                            ". Were masteries created and loaded correctly?");
 
         JoinOrLaunchSession();
-        aim.Select();
+        var result = aim.RecordInteraction(interaction);
+        if (result.IsSuccess)
+        {
+            if (interaction is IPassCausing) TryPass();
```
Luburic, line 88: I don't think that introducing interfaces in this way is a good idea. Essentially, you are using interfaces as a constant property, which is uncommon and therefore confusing. I think that a better option would be to define unserializable boolean properties in the base class that are false and redefine them in the appropriate children.

However, I am also not sure if this is a good idea. Maybe remove it all together and call TryPass and TryComplete all the time?

Luburic (changes requested):
I am not sure what to make of this PR. On the one hand, it does simplify the aggregate. On the other, it distributes some of its logic into the service, making the service and aggregate more coupled _(inapropriate intimacy)._

The main issue is that I am not sure where to place new logic - i.e., the responsibilities are unclear. It seems that the service, aggregate root, and event sourced entity can all generate events. From the perspective of a new programmer joining the team, where should they implement new KCM logic?

Up until this point, it was mostly clear that the service is responsible for loading the aggregate, and activating its main method, and returning the result - as is suitable of a coordinator. That left the KCM messy, but at least the mess was constrained to inside the consistency boundary (i.e., the aggregate).

What do you think?

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/AssessmentItemMastery.cs
milana-todorovic: It should be abstract, I forgot to mark it as such when I added the class. 

I think it makes sense to have a <code>When</code> method for it though, because it makes sense for KCM to apply any events that represent interactions with assessment items the same way - just forward them to the appropriate AIM, and if the mastery estimate for the assessment item changed, recalculate the mastery estimate for the knowledge component as well. The only way I see this changing in the future is if we introduce new mastery indicators which would also need to be updated. 

A more expressive name for <code>AssessmentItemEvent</code> might be useful though, since its meaning is something more like 'the learner interacted with the assessment item', but I'm not sure how to convey that without making it really long.

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
milana-todorovic: Always calling <code>TryPass</code> and <code>TryComplete</code> makes a lot of sense to me theoretically, which is why it was my initial approach as well. I removed the interfaces in 637dbafa2b0736a8b80ca0fe3ade7270792b82d9. I don't actually think it should be an issue in practice either, since calculations performed in these methods aren't that complicated, so trying to mark events which might cause pass or completion is probably just more trouble than it's worth.

src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
milana-todorovic: I don't see how this makes the service smarter, it just fails if the call to <code>RecordAssessmentItemInteraction</code> did?

milana-todorovic (commented):
I do agree that responsibilities are still unclear, but I think that this has been an issue for a while, and that the changes to KCM proposed here actually make it better than it was and not worse. Essentially, I don't think the mess was previously contained within KCM boundaries at all. In this review I went over the changes and marked places that I think actually improve the situation, and places that I think still need to be changed. 

For clarity, here's an overview of all relevant components and how I would describe their responsibilities (as well as their relation to DDD tactical patterns):

- The Knowledge Component Mastery aggregate, which is responsible for tracking a learner's mastery of a knowledge component. This responsibility is achieved by calculating various mastery indicators, and I described the types of commands needed for that in https://github.com/Clean-CaDET/tutor/pull/62#discussion_r981027640, along with which queries should be achieved through this aggregate. This aggregate is event sourced, but that's just an implementation detail which doesn't matter outside of the aggregate (except in the implementation of its repository). It consists of three entities: 
   - <code>SessionTracker</code>, which is responsible for tracking when interaction with a knowledge component is happening (i.e. when mastery updates can happen);
   - <code>AssessmentItemMastery</code>, which is responsible for tracking the learner's mastery of an assessment item within the knowledge component. This is achieved by calculating mastery indicators related to the assessment item, through commands that are all essentially just 'update mastery in accordance with an interaction the learner had with the assessment item';
   - <code>KnowledgeComponentMastery</code>, which is the root and therefore takes on the same responsibility the entire aggregate has. Since some work is delegated to other entities, the work that this class is still directly responsible for is coordinating the entities, and updating the mastery indicators related to the entire knowledge component. 
        
   Other than these entities, the aggregate does contain a reference to the <code>KnowledgeComponent</code> entity, but that entity isn't part of the aggregate. I didn't remove the reference since some data from it like the knowledge component's name needs to be read from the KCM aggregate in some cases. I would very much like to replace it by just inserting the necessary data directly into the <code>KnowledgeComponentMastery</code>, but I'm not sure how that would be mapped with EF Core (EF Core offers computed columns which means that properties can be calculated on fetch, but I don't think you can reference columns in a different table for that). 
- The Knowledge Component aggregate, which is responsible for managing part of the organizational structure of knowledge and educational content (specifically which knowledge components, instructional items, and other knowledge components are contained within a knowledge component). It currently doesn't have any command methods (if it did they would presumably be related to modifications of the structure), but it is used for queries related to the structure it manages. I would actually say that the only entity in this aggregate is <code>KnowledgeComponent</code>. It does contain references to <code>InstructionalItem</code> and <code>AssessmentItem</code> but I don't think they're part of this aggregate since they don't actually maintain any invariants or perform any work together, and there are use cases where methods on <code>AssessmentItem</code> need to be invoked directly, which would mean it's treated as the root of its own aggregate.
- The Assessment Item aggregate, which is responsible for evaluating answers given to an assessment item (This would be considered a query rather than a command, right? It seems a bit unintuitive because of the complexity but I feel like it is definitely a query), finding applicable hints, and such. I can't think of really good phrasing but it's definitely all part of one responsibility. It consists of the <code>AssessmentItem</code> entity.
- The <code>IAssessmentItemSelector</code> domain service, which is responsible for selecting an appropriate assessment item for a learner (Similar to AI answer evaluation, I would also consider this a query). This is currently placed inside the Domain Overlay module but I don't actually think that's correct, since it doesn't track progress, it performs instruction. I think this should be in its own module, probably named Instruction or something similar - maybe within the Learner module, but maybe even completely separate.
- The application services, which coordinate the domain models to perform use cases by loading the necessary elements, invoking their methods, and saving them if needed. For use cases related to the learner's interactions with a knowledge component, one complication is that coordination between two separate domain models will always be needed - either Domain (i.e. Knowledge - I would generally probably rename it to this for clarity) or Instruction, and Domain Overlay (i.e. Mastery or Progress or something like that? Maybe one of those would be a clearer name for it after all). I'm not sure if this kind of coordination is too complicated for application services, or which module is actually the appropriate place for the application services in this case, but the methods for these use cases actually aren't particularly complicated, they contain the same basic steps:
   - Prepare all necessary elements of domain models - this would always be <code>KnowledgeComponentMastery</code>, and either <code>KnowledgeComponent</code>, <code>AssessmentItem</code>, or <code>IAssessmentItemSelector</code>;
   - Invoke the appropriate method on <code>KnowledgeComponent</code>, <code>AssessmentItem</code>, or <code>IAssessmentItemSelector</code>. Currently all of these methods are queries, and I actually think that will always stay the case, since the learner's interactions will presumably never affect knowledge itself in any way, only the learner's mastery of it;
   - Check the result, and fail if necessary;
   - Issue the appropriate command to <code>KnowledgeComponentMastery</code>;
   - Check the second result, and fail if necessary;
   - Save <code>KnowledgeComponentMastery</code>;
   - Return the result.

And I would like to discuss this overview and this PR in general in a sync, both to see if we have any disagreements related to class responsibilities and aggregate structure, and to see how all of this can be further improved (especially the application services and handling of references between aggregates, since that seems like the messiest part right now).

src/Tutor.Core/LearnerModel/DomainOverlay/KnowledgeComponentMasteries/KnowledgeComponentMastery.cs
```diff
@@ -70,40 +68,29 @@ private void JoinOrLaunchSession()
...
+    public Result RecordInstructionalItemSelection()
     {
         JoinOrLaunchSession();
-
         Causes(new InstructionalItemsSelected());
-        return Result.Ok(KnowledgeComponent.InstructionalItems.OrderBy(i => i.Order).ToList());
+        return Result.Ok();
     }
 
-    public Result<int> SelectAssessmentItem(IAssessmentItemSelector assessmentItemSelector)
```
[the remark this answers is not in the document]
Luburic: Agreed. The `IAssessmentItemSelector` is definitely part of the "InstructorModel" which we need to build in the coming months as we will expand our knowledge about the learner (e.g., affective states, behavioral patterns).

src/Tutor.Core/LearnerModel/DomainOverlay/KcMasteryService.cs
```diff
@@ -49,23 +51,31 @@ public Result<List<InstructionalItem>> GetInstructionalItems(int knowledgeCompon
...
             if (kcMastery == null) return Result.Fail("Learner not enrolled in KC: " + knowledgeComponentId);
 
-            var result = kcMastery.SelectInstructionalItems();
+            var result = kcMastery.RecordInstructionalItemSelection();
+            if (result.IsFailed) return result.ToResult<List<InstructionalItem>>();
+
             _kcMasteryRepository.UpdateKcMastery(kcMastery);
 
-            return result;
+            return Result.Ok(kcMastery.KnowledgeComponent.InstructionalItems.OrderBy(i => i.Order).ToList());
```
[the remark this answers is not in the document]
Luburic: As a first step maybe we should just add a method to KC to return its IIs in order?

Luburic (commented):
I agree with most of what you wrote and this is a brilliant analysis. The tricky part for me remains the interaction between the service and the aggregate.

I also agree with your analysis that the _intimacy_ arises from the service using KCM events as a command. This is a tradeoff questions, where we need to have either:
1. Parallel hierarchy structures for the command objects (which is a nightmare to maintain), 
2. Several KCM methods that accept specific parameters and build appropriate events
3. The current intimacy between the service and KCM.

I am not completely convinced that option 3 is better than option 2, but looking at your analysis I am leaning somewhat more towards it. A good way to see which option is better is to understand what would happen in both cases if:
- We add 5 additional methods/events that affect AIM.
- We imagine a different service interacting with the KCM functions that affect AIMs.

While the first bullet is likely to happen, I wouldn't discard the second bullet as a plausible option, especially as we expand the learner model facets.

Luburic:
@lukaDoric Check this PR out and all the comments in it. It will make you a better engineer 🙂

milana-todorovic:
@Luburic Would option 1 with a parallel hierarchy of commands be such a nightmare to maintain though? We kind of automatically discarded it, but I trying adding it in 026fd460dfff42e2e25eae42d558185af48eb452 so we can see what it might look like, and I actually don't think it seems bad at all. Here's a few points in favor of this approach:

- It keeps the main advantage of option 3, which was that new functionality can be simply added to <code>AssessmentItemMastery</code> without having to change <code>KnowledgeComponentMastery</code>, and that <code>KnowledgeComponentMastery</code> no longer contains a bunch of code duplication. If we want to experiment with adding new events just for the sake of storing them without necessarily knowing how they might be used to better estimate mastery in the future (like we're currently doing with <code>SoughtHints</code> and <code>SoughtSolution</code>), we can now do that without changing <code>AssessmentItemMastery</code> or <code>KnowledgeComponentMastery</code> at all, just by adding the event and the command. 
- It also removes the main disadvantage of option 3 - the events used for storage are now contained within boundaries of the KCM aggregate, and their development can focus only on storage needs. 
- The command classes only contain the data necessary to convey the meaning, and access to them can be tightly controlled through constructors and read-only properties, which means that using them from services doesn't have any more potential for error than the method calls from option 2, and has much less potential for error than using events like in option 3. 
- The parallel hierarchies don't actually seem that hard to maintain, mostly because the command hierarchy won't actually need to change often, and because the mapping is contained to the <code>CreateEvent</code> method. I can't really think of any case where adding or changing a command class would be more complicated than adding a method or changing its signature in option 2.

Luburic (commented):
I understand your reasoning and there is some merit to option 1.

However, looking at this code from an outsider perspective (in the sense that I did not write it) it seems there is significant overhead with the addition of these commands. Notably, they seem to be a custom pattern in the sense that they are a variation of the typical command pattern, but the `Execute` method is specialized for `CreateEvent`.

The complexity of creating an event has been migrated from the service to the command class. From what I can tell the command is still part of the application layer, so we did not resolve the issue of the decoupling the app layer from the domain layer, just the issue of decoupling the service from it.

From an understandability perspective I don't think we've enhanced it by introducing additional classes as opposed to having custom methods for each event. You've essentially replaced a custom method with a class and a constructor. The problem I think is that the event creation logic is too simple to justify encapsulation in another class. I do agree, however, that extensibility has improved.

milana-todorovic:
Closed until we can better estimate how bad the issue described in https://github.com/Clean-CaDET/tutor/pull/62#issue-1386911755 will actually be in the future. A new branch will be created for smaller refactoring of KCM which was also discussed on this PR.

PR #78, opened 2023-01-27
Expand SAQ

evaj10:
Adds Feedback and Tolerance to SAQ.

Calculates Levenshtein distance between given and any acceptable answer.
Answer is accepted if distance is less than or equal to Tolerance.

Seed script should be updated.


Luburic (approved):
Looks good.

PR #80, opened 2023-02-09
Notes export

evaj10:
Adds endpoint for notes export to markdown file for authenticated user and given unit id.


Luburic (approved):
Looks good.

PR #82, opened 2023-03-09
Email notifications

evaj10:
Instructor can send email notifications.
Async execution of email sending, to be non-blocking.
Adds sending emails in bulk within one SMTP connection.

Creates ProgressOverviewController for sending emails. Name chosen because we can expand it with endpoint for instructor FE viewing of learner progress overview.

### DIFF
--- a/src/Tutor.Infrastructure/TutorDependencyInjection.cs
+++ b/src/Tutor.Infrastructure/TutorDependencyInjection.cs
@@ -6,6 +6,7 @@ using Tutor.Core.BuildingBlocks.EventSourcing;
 using Tutor.Infrastructure.Database;
 using Tutor.Infrastructure.Database.EventStore.Postgres;
 using Tutor.Infrastructure.Security;
+using Tutor.Infrastructure.Smtp;
 
 namespace Tutor.Infrastructure;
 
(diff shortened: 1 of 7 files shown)


src/Tutor.Infrastructure/TutorDependencyInjection.cs
```diff
@@ -35,4 +39,20 @@ private static string CreateConnectionStringFromEnvironment()
         return
             $"Server={server};Port={port};Database={database};User ID={user};Password={password};Integrated Security={integratedSecurity};Pooling={pooling};";
     }
+
+    private static EmailConfiguration CreateEmailConfigurationFromEnvironment()
+    {
+        var smtpHost = Environment.GetEnvironmentVariable("SMTP_HOST") ?? "smtp.gmail.com";
+        var smtpPort = Environment.GetEnvironmentVariable("SMTP_PORT") ?? "587";
+        var username = Environment.GetEnvironmentVariable("SMTP_USERNAME") ?? "perapera2359@gmail.com";
```
evaj10, line 47: This will be changed to a representative email address.

Luburic (approved):
As a quick and dirty test of the proxy this is ok. However, we should keep in mind (and write it down e.g., on clickup) that there should be an application service (from Core) coordinating the job of extracting learner emails from a repository, their scores, and selecting which email to send (not neccessarily the template, this can be handled by the current EmailService).

To support DIP, we would move the current IEmailService to Core.

Finally, I would rename the IEmailService/EmailService into IEmailSender/EmailSender.

PR #86, opened 2023-03-28
feat: Logging

evaj10:
Implementation of logging on Core/UseCases method level, using DynamicProxy mechanism to implement AOP concepts.

Logs are stored using Docker containers logs. Log rotation is on 500 MB and a maximum of 6 log files are stored.

In src/Tutor.Web/Controllers/BaseApiController.cs, methods for converting FluentResult to ActionResult are implemented. Addition that would simplifies controller classes.
Instead of if block for checking if FluentResult is failed in every controller, one would provides responses for successful and unsuccessful execution, eg. for given result as return object of use case method call:
`return result.CreateResponse(Ok, CreateErrorResponse);`


Luburic (changes requested):
A few comments to address, but overall good work.

src/Tutor.Infrastructure/Interceptors/Logging.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _logger.LogDebug("Calling method {@Class}.{@Method}.", 
+                invocation.TargetType, invocation.Method.Name);
+            invocation.Proceed();
+            try
+            {
+                dynamic result = invocation.ReturnValue;
+                if (result.IsSuccess)
+                {
+                    _logger.LogInformation("Successful execution of method: {@Class}.{@Method}.",
```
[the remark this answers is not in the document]
evaj10: The log before method execution is Debug level, and because of the memory restrictions that were discussed, default level for production was to be Warning. Therefore, that log would not be written. Its purpose was to allow view of method execution time while developing a feature. As this is a rare use case, I will remove it.
Other suggestions are implemented as well.

src/Tutor.Web/appsettings.Development.json
```diff
@@ -1,10 +1,22 @@
...
-      "Default": "Information",
-      "Microsoft": "Warning",
-      "Microsoft.Hosting.Lifetime": "Information"
-    }
+  "Serilog": {
+    "Using": [ "Serilog.Sinks.Console", "Serilog.Sinks.File" ],
+    "MinimumLevel": {
+      "Default": "Warning"
+    },
+    "WriteTo": [
```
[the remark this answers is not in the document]
evaj10: These are not. Checking with Danijel if appsettings.json is read in production, or if everything is configured through Docker.
Should the default logging level for production be set to Warning or Information?

src/Tutor.Web/Extensions/ResultToResponseExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+            Func<ActionResult> onSuccess, Func<List<IError>, ActionResult> onFailure)
+        {
+            return result.IsSuccess ? onSuccess() : onFailure(result.Errors);
+        }
+
+        public static ActionResult CreateResponse<T>(
+            this Result<T> result,
+            Func<T, ActionResult> onSuccess, Func<List<IError>, ActionResult> onFailure)
+        {
+            return result.IsSuccess ? onSuccess(result.Value) : onFailure(result.Errors);
```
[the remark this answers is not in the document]
evaj10: Yes.

I don't see how moving it to BaseApiController would achieve the same effect, but I'm open to examples.

src/Tutor.Infrastructure/Interceptors/Logging.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Microsoft.Extensions.Logging;
+using System.Collections.Generic;
+using Tutor.Core.BuildingBlocks;
+using Castle.DynamicProxy;
+using FluentResults;
+using System;
+
+namespace Tutor.Infrastructure.Interceptors
+{
+    public class Logging : IInterceptor
```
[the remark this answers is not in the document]
Luburic: I would go with the second. The class addresses a technical concern (Logging intercepted information) making the name semantically ok, while still communicating its intent in client code (where we do not easily see the namespace).

src/Tutor.Web/appsettings.Development.json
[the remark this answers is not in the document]
Luburic: I think information would be more standard to understand the broader workings of the system and not only the fail states.

src/Tutor.Web/Extensions/ResultToResponseExtensions.cs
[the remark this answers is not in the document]
Luburic: I was thinking that instead of calling `CreateErrorResponse` from the base class as we do now, we call `CreateResponse` which can handle both success and failure scenarios.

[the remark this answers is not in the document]
evaj10: That we can do, nice catch. Implemented in the latest commit.

src/Tutor.Web/Startup.cs
```diff
@@ -235,22 +237,22 @@ private void SetupServices(IServiceCollection services)
...
 
     private void SetupLearningServices(IServiceCollection services)
     {
-        services.AddScoped<ISessionService, SessionService>();
-        services.AddScoped<IStructureService, StructureService>();
-        services.AddScoped<IStatisticsService, StatisticsService>();
-        services.AddScoped<ISelectionService, SelectionService>();
-        services.AddScoped<IEvaluationService, EvaluationService>();
-        services.AddScoped<IHelpService, HelpService>();
+        services.AddProxiedScoped<ISessionService, SessionService>();
```
Luburic, line 245: Can you briefly explain this change?

src/Tutor.Web/Controllers/Learners/EnrollmentController.cs
```diff
@@ -24,18 +25,16 @@ public class EnrollmentController : BaseApiController
     }
 
     [HttpGet]
-    public ActionResult<List<CourseDto>> GetEnrolledCourses()
+    public ActionResult<PagedResult<CourseDto>> GetEnrolledCourses([FromQuery] int page, [FromQuery] int pageSize)
```
Luburic, line 28: I should have called yesterday to clarify my comment. The thinking for the PagedResult should be:

- If the controller is returning a PagedResult and the client is actually using it meaningfully, we should orient the service and repository to return a PagedResult.
- If it is working with a list, and for some reason it is returning a PagedResult that is always equal to the full list (and content of the DB), just to get the "Count" field, I think we should think about reworking the client to derive the count.

We can always avoid sending the query params to get everything, so it's not a big issue. What do you think?

src/Tutor.Web/Controllers/BaseApiController.cs
```diff
@@ -19,4 +22,52 @@ protected ActionResult CreateErrorResponse(List<IError> errors)
         if (errors.Contains(FailureCode.Conflict)) code = 409;
         return Problem(statusCode: code, detail: string.Join(";", errors));
     }
+
+    protected ActionResult CreateResponse(Result result, Func<ActionResult> onSuccess, Func<List<IError>, ActionResult> onFailure)
```
Luburic, line 26: You should check out the SIMS -> Clean Functions -> Reducing parameter count KC :)

1. We can move the _mapper field to the base class and pass it via constructor.
2. We can set the default success and failure responses since it is always Ok and CreateErrorResponse.

We then have 2 options:

1. Have a single entry point into the base class, that accepts result and determines based on its type which private method to call.
2. Have 3 entry points called `CreateObjectResponse`, `CreateListResponse`, `CreatePagedResponse`.

Luburic (changes requested):
Requires one change and two minor comments to address.

Reviewing these "cross-cutting" changes, even when minor is a pain and it is quite likely that I missed something. We should have split this into two PRs, as we risk introducing a bug in logging because of a controller refactoring change. Something to keep in mind for future PRs.

src/Tutor.Web/Startup.cs
evaj10: Instead of registering services directly to do the ASP.NET service container, we add a Proxy level in between in order to intercept all calls to service methods. Multiple proxies can be added, currently, we have only the Logger.
Sample of execution is given in the picture.

Alt text

src/Tutor.Web/Controllers/Learners/EnrollmentController.cs
evaj10: Yes, I wasn't sure about the approach either.

It is reasonable to have only list response where needed, for example Learner Course navigation in the sidebar. But then, service would have to have to methods, "GetAll" and "GetAllPaged". Because of the implementation of the extension method, which returns everything if "page" and "pageSize" are not present, the methods would be pretty much the same.

Should we have somewhat overlapping endpoints for that? Is there a best practice for API design? Should everything be paged?

evaj10: Also, regarding "GetPaged": because it is async method, where should it be awaited?
Found two ways in the current code:

- in the service: https://github.com/Clean-CaDET/tutor/blob/master/src/Tutor.Core/UseCases/Management/Groups/LearnerGroupService.cs#L27-L29
- in the repo: https://github.com/Clean-CaDET/tutor/blob/master/src/Tutor.Infrastructure/Database/Repositories/Knowledge/CourseDatabaseRepository.cs#L15-L17

src/Tutor.Web/Controllers/BaseApiController.cs
evaj10: The 2. option is a complete (needed) change of perspective. I was thinking of making it one method, generic, the compiler determines as much as it can.
Will take a look the KC, thanks! :D

src/Tutor.Web/Controllers/Learners/EnrollmentController.cs
Luburic: Probably in the service, but pick the more common option.

src/Tutor.Core/BuildingBlocks/FailureCode.cs
```diff
@@ -6,9 +6,12 @@ public static class FailureCode
 {
     public static readonly IError NotEnrolledInUnit = new Error("Not enrolled in unit.");
     public static readonly IError InvalidAssessmentSubmission = new Error("Invalid assessment submission.");
+    public static readonly IError InvalidRefreshToken = new Error("Refresh token is not valid.");
```
Luburic, line 9: Not a fan of this given the FailureCodes are domain-related and not application-related.

Since it's used in one method, I would probably code the `new Error("Refresh token is not valid.")` in the authentication service.

Luburic (approved):
One minor comment and answer to the questions and you can merge after that.

PR #87, opened 2023-03-31
Fix/learning ux

Luburic:
- Resolves a minor issue with event exporting.
- Introduces feature to highlight units where the root KC was mastered by the learner.

### DIFF
--- a/src/Tutor.Web/Controllers/Learners/Learning/StructureController.cs
+++ b/src/Tutor.Web/Controllers/Learners/Learning/StructureController.cs
@@ -23,6 +23,14 @@ public class StructureController : BaseApiController
         _learningStructureService = learningStructureService;
     }
 
+    [HttpPost("units/mastered")]
+    public ActionResult<List<int>> GetMasteredUnitIds([FromBody] int[] unitIds)
+    {
+        var result = _learningStructureService.GetMasteredUnitIds(unitIds, User.LearnerId());
+        if (result.IsFailed) return CreateErrorResponse(result.Errors);
+        return Ok(result.Value);
+    }
+
     [HttpGet("units/{unitId:int}")]
     public ActionResult<KnowledgeUnitDto> GetUnit(int unitId)
     {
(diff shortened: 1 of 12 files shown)


src/Tutor.Web/Controllers/Learners/Learning/StructureController.cs
```diff
@@ -23,6 +23,14 @@ public StructureController(IMapper mapper, IStructureService learningStructureSe
         _learningStructureService = learningStructureService;
     }
 
+    [HttpPost("units/mastered")]
+    public ActionResult<List<int>> GetMasteredUnitIds([FromBody] int[] unitIds)
```
evaj10, line 27: Should this be Http Get?

When asking the question what units have I mastered, we think in context of a course. We could send course id as URL param, with that we can get unit ids. It is more restrictive, but it simplifies the API.

But if there is a need to look at units outside of course context, then this implementation is more reusable.

evaj10 (approved):
Getting used to LINQ trough these PR reviews, real nice work.
Apart from discussion in the comment, approved fully.

src/Tutor.Web/Controllers/Learners/Learning/StructureController.cs
Luburic: This was my original solution, but it required either 3 calls to the database (get enrolled and active units for course + get root KCs + get related masteries).

I thought about compressing the first two calls into a single call, but then I had the problem of positioning the method in the correct repository. Up to now we've used the enrollment repository to work with courses, while the unit and KC repository are mostly focused on that part of the whole structure. I'm also wary of bloating the repositories with too many specific retrieval operations. This can be bypassed with the specification pattern and it is something that we should probably start moving towards.

Regarding the GET vs POST, it is definitely semantically a GET operation and we shouldn't expect more than 10 or 15 unit IDs so I could treat the int array as a query parameter.

evaj10: If your only concern with that aproach are three DB calls, I believe it will not be a bottleneck, as the get enrolled and active units is a fairly simple select.

With the current implementation, if the API user wants to see finished units, he must first send an HTTP request to get the units. This will take more time than a select to DB with an open transaction. 

But if we want to get finished units for more than one course, the current state generic. :)

Makes sense about not joing it into one, because of its position in the repos. In my not so vast past, more smaller queries have had better readability, without a big impact on performance, which I think here is also the case.

PR #88, opened 2023-04-09
Support for SessionPause/SessionContinue.

lukaDoric:
**Change goal:** The goal of this PR is to measure the learner's activity during the session. From now instructor will be able to see the effective time of each session (DurationOfAllSessions - DurationOfAllPauses).
**Required database changes:** KcMasteries table has additional columns (SessionTracker_IsPaused, SessionTracker_DurationOfFinishedPauses, SessionTracker_UnfinishedPauseStart).

### DIFF
--- a/src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
+++ b/src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
@@ -71,6 +71,16 @@ public class KnowledgeComponentMastery : EventSourcedAggregateRoot
         return SessionTracker.Abandon();
     }
 
+    public Result PauseSession()
+    {
+        return SessionTracker.Pause();
+    }
+
+    public Result ContinueSession()
+    {
+        return SessionTracker.Continue();
+    }
+
     private void JoinOrLaunchSession()
     {
         if (!SessionTracker.HasUnfinishedSession) LaunchSession();
--- a/src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
+++ b/src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
@@ -233,6 +233,6 @@ public class KnowledgeComponentMastery : EventSourcedAggregateRoot
 
     private static void When(KnowledgeComponentEvent @event)
     {
-        // No action for AssessmentItemSelected, InstructionalItemsSelected, SolutionRequest, InstructorMessageEvent, SessionTerminated, SessionAbandoned.
+        // No action for AssessmentItemSelected, InstructionalItemsSelected, SolutionRequest, InstructorMessageEvent, and SessionLifecycle events.
     }
 }
--- a/src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
+++ b/src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
@@ -14,10 +14,9 @@ public class SessionTracker : EventSourcedEntity
     public TimeSpan DurationOfFinishedSessions { get; private set; } = new(0);
     public bool HasUnfinishedSession => StartOfUnfinishedSession.HasValue;
 
-    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? LastActivity.Value - StartOfUnfinishedSession.Value : null;
+    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? DateTime.UtcNow - StartOfUnfinishedSession.Value : null;
     public DateTime? StartOfUnfinishedSession { get; private set; }
-    public DateTime? LastActivity { get; private set; }
-    
+
     public bool IsPaused { get; private set; }
     public TimeSpan DurationOfAllPauses => DurationOfFinishedPauses + DurationOfUnfinishedPauses.GetValueOrDefault();
     public TimeSpan DurationOfFinishedPauses { get; private set; } = new(0);
@@ -81,29 +80,22 @@ public class SessionTracker : EventSourcedEntity
     private void When(SessionLaunched @event)
     {
         StartOfUnfinishedSession = @event.TimeStamp;
-        LastActivity = @event.TimeStamp;
         CountOfSessions++;
     }
 
     private void When(SessionTerminated @event)
     {
-        LastActivity = @event.TimeStamp;
-
         DurationOfFinishedSessions += DurationOfUnfinishedSession.GetValueOrDefault();
         StartOfUnfinishedSession = null;
-        LastActivity = null;
     }
 
     private void When(SessionPaused @event)
     {
         LastPause = @event.TimeStamp;
-        
-        LastActivity = LastPause.Value.AddMinutes(-3);
     }
 
     private void When(SessionContinued @event)
     {
-        LastActivity = @event.TimeStamp;
         DurationOfFinishedPauses += @event.TimeStamp - LastPause.Value;
         LastPause = null;
     }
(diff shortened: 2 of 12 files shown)


src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -14,28 +14,59 @@ public class SessionTracker : EventSourcedEntity
...
 
-    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? LastActivity.Value - StartOfUnfinishedSession.Value : null;
+    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? DateTime.UtcNow - StartOfUnfinishedSession.Value : null;
     public DateTime? StartOfUnfinishedSession { get; private set; }
-    public DateTime? LastActivity { get; private set; }
+
+    public bool IsPaused { get; private set; }
+    public TimeSpan DurationOfAllPauses => DurationOfFinishedPauses + DurationOfUnfinishedPauses.GetValueOrDefault();
+    public TimeSpan DurationOfFinishedPauses { get; private set; } = new(0);
+    public TimeSpan? DurationOfUnfinishedPauses => IsPaused ? DateTime.UtcNow - LastPause.Value : null;
```
evaj10, line 23: We should name this DurationOfUnfinishedPause in singular as there cannot be more than one at one time.

src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -14,28 +14,59 @@ public class SessionTracker : EventSourcedEntity
...
+        
+        IsPaused = true;
+        Causes(new SessionPaused());
+        return Result.Ok();
+    }
+
+    public Result Continue()
+    {
+        if (!HasUnfinishedSession) return Result.Fail("No active session to continue.");
+        if (!IsPaused) return Result.Fail("Session is not paused.");
```
evaj10, line 59: Question: Should this check be placed inside Continue method or inside methods that call it? Currently, both places have the same checks.
Opinion: The check is related to the Continue logic, so it should be here. To avoid code duplication, we could do the following: Methods that call Continue don't check if the session is paused, just call the Continue method and return if it is failed.
Current state: Method that call Continue check if IsPaused and only call if correct. Then Continue also checks if  IsPaused.

We still have an if in the method that calls Continue, but we don't double the same check.
Not sure if this is the right approach, @Luburic should comment.

src/Tutor.Core/UseCases/Learning/SessionService.cs
```diff
@@ -50,4 +50,55 @@ public Result TerminateSession(int knowledgeComponentId, int learnerId)
...
+        if (kcMastery == null) return Result.Fail(FailureCode.NotFound);
+
+        var result = kcMastery.AbandonSession();
+        if (result.IsFailed) return result;
+        _knowledgeMasteryRepository.UpdateKcMastery(kcMastery);
+        result = _unitOfWork.Save();
+        if (result.IsFailed) return result;
+
+        return result;
+    }
```
evaj10, line 103: Can we simplify this whole service?
Each method does the same checks, just generates different event.
Should we extract code into two private methods: before and after boiler plate code; and just keep the main logic in the methods?

src/Tutor.Infrastructure/Database/TutorContext.cs
```diff
@@ -191,6 +191,8 @@ private static void ConfigureKnowledgeMastery(ModelBuilder modelBuilder)
         {
             trackerBuilder.Property(tracker => tracker.CountOfSessions).IsRequired().HasDefaultValue(0);
             trackerBuilder.Property(tracker => tracker.DurationOfFinishedSessions).IsRequired().HasDefaultValue(TimeSpan.Zero);
+            trackerBuilder.Property(tracker => tracker.IsPaused).IsRequired().HasDefaultValue(false);
+            trackerBuilder.Property(tracker => tracker.DurationOfFinishedPauses).IsRequired().HasDefaultValue(TimeSpan.Zero);
```
evaj10, line 195: Here we set default to TimeSpan.Zero in DB, but in the SessionTracker line 17 we set it to null. Should we use TimeSpan.Zero there too?
Just curiosity.

evaj10 (commented):
Really good code and completely valid logic. Nicely thought out and tested.
Well done!
Comments are mostly potential code simplifications.

src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
```diff
@@ -71,6 +71,16 @@ public Result AbandonSession()
...
+        return SessionTracker.Pause();
+    }
+
+    public Result ContinueSession()
+    {
+        return SessionTracker.Continue();
+    }
+
     private void JoinOrLaunchSession()
     {
```
Luburic, line 85: Should rephrase the comment at the bottom of this class (in private static void When(KnowledgeComponentEvent @event)) to cover all Session Lifecycle events. Instead of appending the Continue and Pause, set the comment to
`// No action for AssessmentItemSelected, InstructionalItemsSelected, SolutionRequest, InstructorMessageEvent, and SessionLifecycle events.`

src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -14,28 +14,59 @@ public class SessionTracker : EventSourcedEntity
...
-    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? LastActivity.Value - StartOfUnfinishedSession.Value : null;
+    public TimeSpan? DurationOfUnfinishedSession => HasUnfinishedSession ? DateTime.UtcNow - StartOfUnfinishedSession.Value : null;
     public DateTime? StartOfUnfinishedSession { get; private set; }
-    public DateTime? LastActivity { get; private set; }
+
+    public bool IsPaused { get; private set; }
+    public TimeSpan DurationOfAllPauses => DurationOfFinishedPauses + DurationOfUnfinishedPauses.GetValueOrDefault();
+    public TimeSpan DurationOfFinishedPauses { get; private set; } = new(0);
+    public TimeSpan? DurationOfUnfinishedPauses => IsPaused ? DateTime.UtcNow - LastPause.Value : null;
+    public DateTime? LastPause { get; private set; }
```
Luburic, line 24: This name is not accurate. First, it does not model the concept of a "Pause". Next, it does not maintain the last pause time as it is set to null when the pause is over.

Maybe `UnfinishedPauseStart`?

src/Tutor.Web/Mappings/Enrollments/GroupProfile.cs
```diff
@@ -15,7 +15,10 @@ public GroupProfile()
         CreateMap<KnowledgeComponentMastery, KcmProgressDto>()
             .ForMember(dest => dest.DurationOfAllSessionsInMinutes, opt => opt.MapFrom(src =>
                 (src.SessionTracker.DurationOfAllSessions.Hours * 60) +
-                src.SessionTracker.DurationOfAllSessions.Minutes));
+                src.SessionTracker.DurationOfAllSessions.Minutes))
+            .ForMember(dest => dest.DurationOfAllPausesInMinutes, opt => opt.MapFrom(src =>
```
Luburic, line 19: Not sure if this is the correct place for this logic or if it is needed.

Does this give accurate results https://learn.microsoft.com/en-us/dotnet/api/system.timespan.totalminutes?view=net-7.0 ?

src/Tutor.Core/UseCases/Learning/SessionService.cs
Luburic: Yes, good catch.

src/Tutor.Core/Domain/KnowledgeMastery/SessionTracker.cs
Luburic: The final check should be in the `Continue` method to ensure a valid state.

However, the `IsPaused` check in, for example, the `Terminate` method above (line 40) is not a validation rule but a control flow rule. We could remove the conditional expression and just call Continue, knowing it will fail when it needs to, but such code is less expressive.

Luburic (changes requested):
Added a few minor comments and reviewed @evaj10 's comments.

Other than what is mentioned we are missing:
- A brief description of the PR in its header, including any **changes to the database required**. Let's introduce the following headers as a convention for any PR description:
   - "Change Goal" where we briefly describe what the expected outcome of the feature is.
   - "Required Database Changes" and list "None" if none are required.
   - We will expand this convention as necessary.
- Automated tests. I want to avoid dedicating a day once every two months to build a test-focused PR.

Luburic (approved):
Looks good.

PR #91, opened 2023-04-27
Fix/email config

evaj10:
## Change goal

- Handle exception introduced by email proxy configuration
   - Surround with try/catch block
   - Add configuration and error logs to ensure parameters are passed to the Docker container


Luburic (approved):
Looks good.

PR #93, opened 2023-05-06
Feat/learner registration

evaj10:
## Change goal

- Enable special handling of commercial learners
    - New user role added ('LearnerCommercial')
    - New enumeration field to Learner class added ('LearnerType')
    - New field to StakeholderAccountDto added ('LearnerType')
- Handle username uniqueness on registration and bulk registration
    - Return 400 if username already exists on single registration
- Adds tests for login and refresh with checks if adequate role is present
- Adds tests for fetching all learners with checks if LearnerType is present

## Logic changes

- Bug fix: on learner update, username now equals index not email

## Database changes

- New required int column LearnerType with default value 0 in Learner relation

### DIFF
--- a/src/Tutor.Core/Domain/Stakeholders/Learner.cs
+++ b/src/Tutor.Core/Domain/Stakeholders/Learner.cs
@@ -7,4 +7,12 @@ public class Learner : Stakeholder
 {
     public string Index { get; private set; }
     public List<KnowledgeComponentMastery> KnowledgeComponentMasteries { get; private set; } = new();
+    public LearnerType LearnerType { get; private set; }
+}
+
+public enum LearnerType
+{
+    // Should values match UserRole values?
+    Regular,
+    Commercial
 }
--- a/src/Tutor.Core/Domain/Stakeholders/Learner.cs
+++ b/src/Tutor.Core/Domain/Stakeholders/Learner.cs
@@ -12,7 +12,6 @@ public class Learner : Stakeholder
 
 public enum LearnerType
 {
-    // Should values match UserRole values?
     Regular,
     Commercial
 }
--- a/src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
+++ b/src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
@@ -1,4 +1,6 @@
 ﻿using FluentResults;
+using System;
+using System.Text.RegularExpressions;
 using Tutor.Core.BuildingBlocks;
 using Tutor.Core.BuildingBlocks.Generics;
 using Tutor.Core.Domain.Stakeholders;
@@ -9,14 +11,18 @@ namespace Tutor.Core.UseCases.Management.Stakeholders;
 public class StakeholderService<T> : CrudService<T>, IStakeholderService<T> where T : Stakeholder
 {
     private readonly IUserRepository _userRepository;
+    private static readonly Regex _whitespacesRegex = new Regex(@"\s+");
 
     public StakeholderService(ICrudRepository<T> crudRepository, IUnitOfWork unitOfWork, IUserRepository userRepository) 
         : base(crudRepository, unitOfWork)
     {
         _userRepository = userRepository;
     }
-    public Result<T> Register(T entity, string username, string password, UserRole role)
+    public Result<T> Register(T entity, string username, string password, string userType)
     {
+        var result = GetRole(userType, out UserRole role);
+        if (result.IsFailed) return result;
+
         UnitOfWork.BeginTransaction();
 
         var existingUser = _userRepository.GetByName(username);
@@ -24,7 +30,7 @@ public class StakeholderService<T> : CrudService<T>, IStakeholderService<T> wher
             return Result.Fail(FailureCode.DuplicateUsername);
 
         var user = _userRepository.Register(username, password, role);
-        var result = UnitOfWork.Save();
+        result = UnitOfWork.Save();
         if (result.IsFailed)
         {
             UnitOfWork.Rollback();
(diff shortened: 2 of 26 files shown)


src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
```diff
@@ -1,26 +1,37 @@
...
+using System.Text.RegularExpressions;
 using Tutor.Core.BuildingBlocks;
 using Tutor.Core.BuildingBlocks.Generics;
 using Tutor.Core.Domain.Stakeholders;
 using Tutor.Core.Domain.Stakeholders.RepositoryInterfaces;
 
 namespace Tutor.Core.UseCases.Management.Stakeholders;
 
-public class StakeholderService<T> : CrudService<T>, IStakeholderService<T> where T : Stakeholder
+public partial class StakeholderService<T> : CrudService<T>, IStakeholderService<T> where T : Stakeholder
```
Luburic, line 11: No need to modify the class in this way. I think you can achieve this with the `Regex.Replace`.

However, I think this logic needs to be reworked in general, so we should schedule a quick call to cleanup the domain and user interplay.

src/Tutor.Web/Controllers/Administrators/Stakeholders/InstructorController.cs
```diff
@@ -29,7 +29,7 @@ public ActionResult<PagedResult<StakeholderAccountDto>> GetAll([FromQuery] int p
     [HttpPost]
     public ActionResult<StakeholderAccountDto> Register([FromBody] StakeholderAccountDto stakeholderAccount)
     {
-        var result = _instructorService.Register(_mapper.Map<Instructor>(stakeholderAccount), stakeholderAccount.Email, stakeholderAccount.Password, UserRole.Instructor);
+        var result = _instructorService.Register(_mapper.Map<Instructor>(stakeholderAccount), stakeholderAccount.Email, stakeholderAccount.Password, "Instructor");
```
Luburic, line 32: UserRole.Instructor.ToString()?

Luburic (changes requested):
There are a few minor and one major thing to change before another round of reviews. Ideally we can fit in a call tomorrow (Friday) to sketch a sequence diagram and see how we can improve the design.

src/Tutor.Core/Domain/Stakeholders/Learner.cs
```diff
@@ -7,4 +7,12 @@ public class Learner : Stakeholder
 {
     public string Index { get; private set; }
     public List<KnowledgeComponentMastery> KnowledgeComponentMasteries { get; private set; } = new();
+    public LearnerType LearnerType { get; private set; }
+}
+
+public enum LearnerType
+{
+    // Should values match UserRole values?
```
Luburic, line 15: These are ok labels for now.

src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
```diff
@@ -54,15 +57,15 @@ public Result<T> Archive(int id, bool archive)
         return Result.Ok(stakeholder);
     }
 
-    public override Result<T> Update(T entity)
+    public Result<T> Update(T entity, string username)
```
Luburic, line 60: This "special case" seems like a code smell stemming from the Stakeholder hierarchy. Not a trivial thing to fix so we can leave it for now, but we should pay attention to this part of the code down the line and maybe improve it for the summer redesign.

src/Tutor.Web/Controllers/Administrators/Stakeholders/LearnerController.cs
```diff
@@ -85,4 +102,35 @@ public ActionResult<PagedResult<CourseDto>> GetEnrolledCourses(int id, [FromQuer
...
+        {
+            Error rootError = new Error(e.Message).CausedBy(e);
+            return Result.Fail("Unsupported learner type.").WithError(rootError);
+        }
+        return Result.Ok(learner);
+    }
+
+    private Result<List<Learner>> MapLearners(List<StakeholderAccountDto> stakeholderAccounts)
+    {
+        if (stakeholderAccounts.Select(s => s.LearnerType).Distinct().Count() > 1)
```
Luburic, line 123: This whole logic seems like it belongs in the service and should be moved there.

Luburic (changes requested):
One thing to address and you can merge.

src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
evaj10: Agreed on the hierarchy problems. Should this change to update logic be included in the PR?
If not, updating learner will changes its credentials from index to email.

src/Tutor.Web/Controllers/Administrators/Stakeholders/LearnerController.cs
evaj10: Logic is related to DTO mappings and handles mapping exceptions, not sure how to migrate to service level. Any suggestions?

Luburic: How do we handle mapping exceptions in general?

evaj10: I don't think we do. This exception seems to happen only when trying to map to enum value if enum value is invalid. AutoMapper handles invalid type conversion for other values as BadRequests.

src/Tutor.Core/UseCases/Management/Stakeholders/StakeholderService.cs
evaj10: We could if the hierarchy for Learner/LearnerCommercial were separate from Instructor which does not have the index field.

src/Tutor.Web/Controllers/Administrators/Stakeholders/LearnerController.cs
Luburic: Hmm, my guess is that this is not the only place that deals with enums at the endpoint level. Was this a fix for the 500 error found by @lukaDoric ?

If this code is necessary we can close the PR and open the one for dev into master that we talked about.

PR #97, opened 2023-05-16
fix: subtract 2 minutes from UnfinishedPauseStart.

lukaDoric:


src/Tutor.Core/Domain/KnowledgeMastery/KnowledgeComponentMastery.cs
```diff
@@ -176,7 +177,11 @@ public Result RecordInstructorMessage(string message)
     protected override void Apply(DomainEvent @event)
     {
         if (@event is not KnowledgeComponentEvent kcEvent) throw new EventSourcingException("Unexpected event type: " + @event.GetType());
-
+        if (kcEvent is SessionPaused)
```
Luburic, line 180: Don't add a special case here. Modify the timestamp at the place where the event is created.

Luburic (changes requested):
Logic is poorly placed and needs to improve.

Luburic (approved):
Looks good.

PR #99, opened 2023-05-28
Fix/delete learner

evaj10:
## Goal

Cascade delete UnitEnrollments and GroupMemberships when deleting Learner.

## Tests

- Adds tests for delete Learner.
- Adds tests for potential update Stakeholder bug cause.


Luburic (approved):
Looks good. We should plan to solve the stakeholder issue in the summer redesign.

PR #100, opened 2023-05-29
Development

Luburic:
## Goal
Improves course and enrollment lifecycle controls and introduces some cleanup.

## List of changes
- "Enrollment" is considered active when it has the appropriate status and after it's start time.
- Instructors can "Bulk unenroll".
- Instructors can select Date for "Enrollment" start.
- Learners cannot access archived courses.
- Archiving a course sets all enrollments to hidden.
- Removes deprecated excel data importer and related library.
- Adds cascade delete UnitEnrollments and GroupMemberships when deleting Learner.
- Minor migration of logic from enrollment repository to upper layers.

## Required database changes
None.


evaj10 (commented):
All looking good, nice cleanup!
One thing to potentially address in this PR: instructors can access archived courses and change enrollment for users. When fetching courses for instructors, we should filter only active ones.

Luburic:
> All looking good, nice cleanup! One thing to potentially address in this PR: instructors can access archived courses and change enrollment for users. When fetching courses for instructors, we should filter only active ones.

I think that instructors should be able to view archived courses (e.g., to check how a particular learner did, to view a specific assessment item), but we can constrain what can be done on an archived course (e.g., ban authoring, enrollment mngmt...). I have created a ticket for this. https://app.clickup.com/t/861mwa5mj

PR #105, opened 2023-08-11
Language model integration

evaj10:
## Change goal

- Introduce language model integration for learing
    - New module added -> *LanguageModelConversations*
    - Learner can interact with a language model through open-ended conversation and predefined actions:
        - Summarize
        - Generate Questions
        - Extract Keywords
        - Generate Similar
    - Learner has a predefined amount of tokens allocated for the conversations
        - In case of insufficient funds, 402 status code is returned
    - Module handles Summarization storage to lower the amount of tokens spent on static content
    - Learner cannot converse about content he has not been given access (units for which enrollment does not exist)

## Database changes

- New schema: *languageModelConversations* with three tables:
    - Conversations
    - TokenWallets
    - Summarizations

## Diagrams
Open chat - existing and non existing chat
Send message


evaj10:
The code has two questions regarding the implementation.

Luburic:
Closing this stale PR since we are using another approach to integrate LLMs that is currently finalizing development.

PR #107, opened 2023-11-22
Implemented modal for students emotions annotation

katarina-glorija:
Added a modal window that pops up every 20 minutes and asks a student to annotate their emotion.


Luburic (approved):
Resolved minor issues and merging to development.

PR #127, opened 2024-07-25
Catching LearningTask events.

lukaDoric:
## Change goal

The goal of this PR was to enable the backend to capture the following events:

TaskOpened
StepOpened
SubmissionOpened
GuidanceOpened
ExampleOpened
ExampleVideoPlayed
ExampleVideoPaused
ExampleVideoFinished
StepSubmitted
TaskCompleted

## Database changes

Add Events table to learningTasks using the following script.

```sql
CREATE TABLE "learningTasks"."Events"
(
    "Id" integer NOT NULL GENERATED BY DEFAULT AS IDENTITY ( INCREMENT 1 START 1 MINVALUE 1 MAXVALUE 2147483647 CACHE 1 ),
    "AggregateType" text COLLATE pg_catalog."default" NOT NULL,
    "AggregateId" integer NOT NULL,
    "TimeStamp" timestamp with time zone NOT NULL,
    "DomainEvent" jsonb NOT NULL,
    CONSTRAINT "PK_Events" PRIMARY KEY ("Id")
)

TABLESPACE pg_default;

CREATE INDEX "IX_Events_TimeStamp"
    ON "learningTasks"."Events" USING btree
    ("TimeStamp" ASC NULLS LAST)
    TABLESPACE pg_default;
```


src/Modules/LearningTasks/Tutor.LearningTasks.Core/Domain/LearningTaskProgress/Events/TaskProgressEvents/StepOpened.cs
```diff
@@ -0,0 +1,12 @@
+﻿
```
Luburic, line 1: Many events have these empty lines. Remove them.

src/Modules/LearningTasks/Tutor.LearningTasks.Infrastructure/Tutor.LearningTasks.Infrastructure.csproj
```diff
@@ -6,6 +6,10 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="Dahomey.Json" Version="1.12.2" />
```
Luburic, line 10: If you manage to migrate most of the configuration to `BuildingBlocks.Infrastructure`, check if you can remove this dependency.

src/Modules/KnowledgeComponents/Tutor.KnowledgeComponents.Infrastructure/KnowledgeComponentsStartup.cs
```diff
@@ -91,8 +91,8 @@ private static void SetupInfrastructure(IServiceCollection services)
 
         services.AddScoped<IKnowledgeMasteryRepository, KnowledgeMasteryDatabaseRepository>();
 
-        services.AddScoped<IEventStore, PostgresStore>();
-        services.AddSingleton<IEventSerializer>(new DefaultEventSerializer(EventSerializationConfiguration.EventRelatedTypes));
+        services.AddScoped<IKnowledgeComponentEventStore, PostgresStore>();
```
Luburic, line 94: It seems that both `IEventStore` and `IEventSerializer` can remain in `BuildingBlocks.Infrastructure` if we make the interfaces parametrized.

Look to the generic repositories:
- Generic database repo https://github.com/Clean-CaDET/tutor/blob/master/src/BuildingBlocks/Tutor.BuildingBlocks.Infrastructure/Database/CrudDatabaseRepository.cs
- Startup configuration https://github.com/Clean-CaDET/tutor/blob/master/src/Modules/Courses/Tutor.Courses.Infrastructure/CoursesStartup.cs#L56

src/Modules/LearningTasks/Tutor.LearningTasks.Core/Domain/LearningTaskProgress/TaskProgress.cs
```diff
@@ -35,16 +38,80 @@ public void SubmitAnswer(int stepId, string answer)
...
+
+        kcEvent.LearningTaskId = LearningTaskId;
+        kcEvent.LearnerId = LearnerId;
+
+        When((dynamic)kcEvent);
+    }
+
+    private static void When(TaskProgresskEvent @event)
+    {
+        // No additional actions are required for all events.
```
Luburic, line 114: The `When` methods should change the state of the aggregate. At a minimum this means that `When(StepOpened)` and `When(StepSubmitted` occur we should invoke logic that is currently misplaced in the methods `SubmitAnswer` and `ViewStep`.

To summarize:
- In an event-source aggregate, methods called by services generate events,
- Event-handling methods change the aggregate's state.

src/Modules/LearningTasks/Tutor.LearningTasks.Core/UseCases/Learning/TaskProgressService.cs
```diff
@@ -55,6 +62,8 @@ public Result<TaskProgressDto> ViewStep(int unitId, int id, int stepId, int lear
 
         var taskProgress = result.Value;
         taskProgress.ViewStep(stepId);
+        _progressRepository.UpdateEvents(taskProgress);
+
         return Update(taskProgress);
```
Luburic, line 67: Will this generate two events (one for the `UpdateEvents` and another for `Update` in line 67?

src/Modules/LearningTasks/Tutor.LearningTasks.Core/UseCases/Learning/TaskProgressService.cs
```diff
@@ -66,9 +75,89 @@ public Result<TaskProgressDto> SubmitAnswer(int unitId, int id, StepProgressDto
 
         var taskProgress = result.Value;
         taskProgress.SubmitAnswer(stepProgress.StepId, stepProgress.Answer!);
+        _progressRepository.UpdateEvents(taskProgress);
+
         return Update(taskProgress);
     }
 
+    public Result OpenSubmission(int unitId, int id, int stepId, int learnerId)
```
Luburic, line 83: I think all of these non-state changing methods can be simplified to reduce code duplication. This might not work, as I am writing from memory instead of in an IDE:
```csharp
public Result PlayExampleVideo(int unitId, int id, int stepId, int learnerId)
{
    return HandleNonStateChangingEvent(unitId, id, stepId, learnerId, (taskProgress, step) => taskProgress.PlayExampleVideo(step));
}

public Result PauseExampleVideo(int unitId, int id, int stepId, int learnerId)
{
    return HandleNonStateChangingEvent(unitId, id, stepId, learnerId, (taskProgress, step) => taskProgress.PauseExampleVideo(step));
}

public Result FinishExampleVideo(int unitId, int id, int stepId, int learnerId)
{
    return HandleNonStateChangingEvent(unitId, id, stepId, learnerId, (taskProgress, step) => taskProgress.FinishExampleVideo(step));
}

private Result HandleNonStateChangingEvent(int unitId, int id, int stepId, int learnerId, Action<TaskProgress, int> action)
{
    var result = GetTaskProgress(unitId, learnerId, id);
    if (result.IsFailed)
        return result.ToResult();

    var taskProgress = result.Value;
    action(taskProgress, stepId);
    _progressRepository.UpdateEvents(taskProgress);

    return UnitOfWork.Save();
}
```

Luburic (changes requested):
Comments inline.

src/Modules/LearningTasks/Tutor.LearningTasks.Core/UseCases/Learning/TaskProgressService.cs
Luburic: Check if this method specifically removes any old events from existence. I do not think this will be the case, but better safe than sorry.

src/Modules/LearningTasks/Tutor.LearningTasks.Core/UseCases/Learning/TaskProgressService.cs
lukaDoric: Nice

src/Modules/LearningTasks/Tutor.LearningTasks.Core/UseCases/Learning/TaskProgressService.cs
lukaDoric: It does not save any events without _progressRepository.UpdateEvents(taskProgress);.

src/Modules/LearningTasks/Tutor.LearningTasks.Tests/Integration/Learning/TaskProgressTests.cs
```diff
@@ -250,6 +250,78 @@ public void Fails_to_submit_answer_for_missing_task()
...
+    public void FinishExampleVideo()
+    {
+        using var scope = Factory.Services.CreateScope();
+        var controller = CreateController(scope);
+        var dbContext = scope.ServiceProvider.GetRequiredService<LearningTasksContext>();
+
+        var result = controller.FinishExampleVideo(-2, -2, -4, "videoUrl");
+
+        result.ShouldBeOfType<OkResult>();
+    }
```
Luburic, line 323: A Test Suite with 300+ lines of code (or really 200+) should be a signal (code smell) for refactoring.

src/Tutor.API/Controllers/Instructor/Analysis/EventsController.cs
```diff
@@ -2,20 +2,21 @@
...
+using Tutor.BuildingBlocks.Infrastructure.Database.EventStore.DefaultEventSerializer;
 using Tutor.KnowledgeComponents.Infrastructure.Database.EventStore;
-using Tutor.KnowledgeComponents.Infrastructure.Database.EventStore.DefaultEventSerializer;
 
 namespace Tutor.API.Controllers.Instructor.Analysis;
 
 [Authorize(Policy = "instructorPolicy")]
 [Route("api/analysis/knowledge-components/events")]
-public class EventsController : BaseApiController
+public class EventsController<TEvent> : BaseApiController where TEvent : DomainEvent
```
Luburic, line 14: No need for <TEvent> in the controller. I guess it was caused by a lack of focus, but we cannot parametrize all the layers of our application when we only need a paramterized infrastructure service.

Luburic (commented):
Apart from the issues in the comments below, the PR text was not updated to reflect the required database changes and the TestData SQL script was not updated to delete events.

I will solve the listed issues and resolve the Sonar issues as well.

PR #129, opened 2024-08-08
Fixed bug in SessionTracker

lukaDoric:
## Change goal

The goal of this PR was to fix bugs within the SessionTracker that occur when the computer is shut down or goes into sleep mode.

## Database changes

ALTER TABLE IF EXISTS "knowledgeComponents"."KcMasteries"
    ADD COLUMN "SessionTracker_LastActivity" timestamp with time zone;


src/Modules/KnowledgeComponents/Tutor.KnowledgeComponents.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -44,19 +54,28 @@ public Result Terminate()
 
     public Result Pause()
     {
-        if (!HasUnfinishedSession) Launch();
+        if (!HasUnfinishedSession) return Result.Fail("No active session to pause.");
```
Luburic, line 57: In the scenario where people have left their browser running and shutdown their PC - the pause event triggers as soon as the PC is started? Will it have an active session?

What if they opened the Tutor in another app (e.g., mobile) in the meantime? The other session should activate, abandoning this one? Verify that such a scenario cannot lead to errors in events.

src/Modules/KnowledgeComponents/Tutor.KnowledgeComponents.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -44,19 +54,28 @@ public Result Terminate()
...
         if (IsPaused) return Result.Fail("Session is already paused.");
         
         IsPaused = true;
```
Luburic, line 60: State changes should be in the event handling (When method). `IsPaused = true;` is a state change.

src/Modules/KnowledgeComponents/Tutor.KnowledgeComponents.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -44,19 +54,28 @@ public Result Terminate()
...
         return Result.Ok();
     }
 
     public Result Continue()
     {
-        if (!HasUnfinishedSession) Launch();
+        if (!HasUnfinishedSession) return Result.Fail("No active session to continue.");
         if (!IsPaused) return Result.Fail("Session is not paused.");
             
         IsPaused = false;
```
Luburic, line 81: State change.

src/Modules/KnowledgeComponents/Tutor.KnowledgeComponents.Core/Domain/KnowledgeMastery/SessionTracker.cs
```diff
@@ -27,7 +29,15 @@ public void Launch()
         if (HasUnfinishedSession)
         {
             if (IsPaused) Continue();
-            Causes(new SessionAbandoned());
+
+            var sessionAbandoned = new SessionAbandoned();
+
+            if (LastActivity.HasValue && (sessionAbandoned.TimeStamp - LastActivity.Value).TotalHours > 1)
```
Luburic, line 35: There seems to be abug here. Line 31 `if (IsPaused) Continue();` will generate a session continued event that has the current datetime, while the session abandoned might be some time in the past.

We should create the continued event here as well, making the state change comment below even more important.

Luburic (changes requested):
Several fixes needed. Before you merge ensure that the LastActivity is correctly set for all KC events.

Luburic (approved):
Looks good.

